PR #1, opened 2022-03-20
make it possible to select the target

thvdveld:
Still need to find a way for the `hex_literal` replacement.


thvdveld:
To compile it for `cc2538` you need to compile with:
`cargo build --target thumbv7m-none-eabi --no-default-features --features cc2538`

PR #3, opened 2022-03-25
Add support for accelerators

thvdveld:


malishav:
@thvdveld thanks for working on this. The PR is still marked as draft. Let me know once I should go through it. One thing that caught my eye is that cc2538-specific files like `flash.sh`, `gdb.sh`, etc, should be under `cc2538-edhoc`

thvdveld:
@malishav Yes indeed, it was marked as draft, because I was not happy with the current solution to test the implementation on the CC2538. Currently the problem is that it lives in another crate, meaning that I had to change the visibility of the functions to public to be able to use them in the cc2538-crate. I have no idea how to even fix this.

malishav:
Would it be so bad if we don't run the whole test suite with generic functions on the microcontroller, but rather have a dedicated test suite for the microcontrollers with only the unit crypto functions executing in test mode? In practice, that would mean calling functions like `test_sha256_digest` or `test_p256_ecdh` but not "higher-layer" functions like `test_parse_message_2` or `test_decrypt_ciphertext_2`. What do you think?

thvdveld:
I agree! I removed the `pub` keyword. It should be like before now.

malishav:
Ok, I just went through the new tree. One thing that doesn't make sense to me is that there is a `cc2538.rs` file under `edhoc/src/accelerator/ and then there is a separate `edhoc-cc2538` directory under root. I guess the purpose is that `edhoc-cc2538` serves as an example how to use this on cc2538? In that case, shouldn't `edhoc-cc2538` live under `edhoc/examples` as a separate crate with dependency on the root edhoc crate? Is it possible to have the accelerator also implemented in that folder?

thvdveld:
Yes that is possible if that is how you want it! I just thought that you also wanted to have the acceleration of CC2538 specific things in the EDHOC crate (which is not really the point of EDHOC I think). So indeed, it should be moved. Will do that as well!

thvdveld:
Your understanding is correct! For now we leave `cc2538-edhoc` in examples I guess. I don't think it belongs there since it is more than an example: it is an actual implementation of EDHOC for the CC2538. However, it is an example of using an accelerated version of EDHOC.

malishav:
It looks almost ready to fly, but when I do `cargo test` hoping to launch the test suite on native, it complains about some cortex m macros. Could you check on your side?

thvdveld:
Fixed it now. I still had some `dev-dependencies`.

thvdveld:
If running `cargo build --target thumbv7m-none-eabi --no-default-features --features cc2538`, then build fails, but that is something I need to fix in cc2538-hal.

But running `cargo build --target-thumbv7m-none-eabi` in `examples/cc2538-edhoc` should be fine.

malishav:
thanks for working on this!

thvdveld:
No problem at all!

`cargo build --target thumbv7m-none-eabi --no-default-features --features cc2538` should now build fine (after running `cargo update`).

PR #19, opened 2022-11-08
use cargo's `[patch]` to point to hacspec aesccm branch

kaspar030:


malishav:
Thanks for providing this PR, this all looks good to me.

PR #23, opened 2022-11-08
move examples out of lib/

kaspar030:
This moves the coap and cc2538 examples out of the lib/ crate into their own workspace crates.

The main benefit is that now edhoc-rs doesn't have to specify any coap dependencies anymore.

Another piggybacked commit removes edhoc-rs' "hex-literal" dependency, which seems to have been superseded by "hexlit".


malishav:
I like how this simplifies the library dependencies, thansk for providing the PR!

PR #24, opened 2022-11-09
Add plain no std example

kaspar030:
This PR adds a no_std example with working qemu Cortex-M3 target configuration and runner config.
Currently it makes only very limited use of edhoc-rs, but should be a good base for more no_std testing.

This was previously at https://github.com/kaspar030/edhoc-rs-no_std. The workspace level .cargo/config.toml previously prevented correct building from inside this (edhoc-rs) repo, which I didn't realize, so I had put it somewhere else.
Without that file (which is IMO not needed there), the example can be moved here.


examples/edhoc-rs-no_std/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+static-alloc = { version = "0.2.3" }
+
+# cortex-m specifics
+cortex-m = "0.7.4"
+cortex-m-rt = "0.7.1"
+cortex-m-semihosting = "0.5.0"
+panic-semihosting = { version = "0.6.0", features = ["exit"] }
+
+[patch.crates-io]
+hacspec-lib = { git = "https://github.com/malishav/hacspec", branch = "aesccm" }
```
malishav, line 21: Is this necessary given that it is already present in the workspace `Cargo.toml`?

examples/edhoc-rs-no_std/build.rs
```diff
@@ -0,0 +1,31 @@
...
+//! Cargo re-run the build script whenever `memory.x` is changed,
+//! updating `memory.x` ensures a rebuild of the application with the
+//! new memory settings.
+
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::path::PathBuf;
+
+fn main() {
```
malishav, line 16: I am likely missing something obvious but I don't understand how is this `build.rs` script invoked during build? Does cargo automatically look for `build.rs` in its path?

examples/edhoc-rs-no_std/Cargo.toml
kaspar030: It is if this example is excluded from the main workspace.
But I think I like better the solution in #25 there the crate is part of the workspace, but not part of the `default-members`.
I'll wait for #25 and see if that method works here, too!

examples/edhoc-rs-no_std/build.rs
kaspar030: Yes, cargo automatically builds & runs build.rs!

kaspar030: Here's more info: https://doc.rust-lang.org/cargo/reference/build-scripts.html

examples/edhoc-rs-no_std/Cargo.toml
kaspar030: yes, that works. I've implemented it the same way (and removed this `[patch.crates-io]`).

kaspar030:
- rebased
- `[patch]` removed
- added github workflow for building the test. broke my promise about the matrix. :)

PR #26, opened 2022-11-09
examples/coap: some cleanup

kaspar030:
- use Cargo's binary auto detection (every `bin/<name>.rs` becomes a `[[bin]]` with `<name>` automatically, no need for Cargo.toml entries)
- fix all compilation warnings in coapserver.rs


malishav:
Just tested locally, all seems OK!

PR #27, opened 2022-11-09
crypto: add initial PSA work

kaspar030:
This adds a `edhoc-crypto-psa` package to the build system.
The primitives are not implemented yet!

edhoc-rs can be tested with this backend with:

`cargo test --package edhoc-rs --no-default-features --features "hacspec-psa"`

... but it fails due to missing implementations.

This PR is based on #25, which should be reviewed and merged first.


kaspar030:
> Could you rebase this one now that #25 is in?

rebased. I'm having some trouble building psa-crypto for no_std. :(

PR #30, opened 2022-11-10
crypto: add psa backend (stubs!)

kaspar030:
This adds a edhoc-crypto-psa package to the build system.
The primitives are not implemented yet!

edhoc-rs can be tested with this backend with:

cargo test --package edhoc-rs --no-default-features --features "hacspec-psa"

... but it fails due to missing implementations.

This PR is based on https://github.com/openwsn-berkeley/edhoc-rs/pull/25, which should be reviewed and merged first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
   "crypto",
   "crypto/edhoc-crypto-cc2538",
   "crypto/edhoc-crypto-hacspec",
+  "crypto/edhoc-crypto-psa",
   "examples/coap",
   "examples/edhoc-rs-no_std",
 ]
@@ -17,6 +18,7 @@ default-members = [
   "hacspec",
   "crypto",
   "crypto/edhoc-crypto-hacspec",
+  "crypto/edhoc-crypto-psa",
   "examples/coap",
 ]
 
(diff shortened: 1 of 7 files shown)


malishav:
Just a note that I got `psa-crypto` to compile on MAC OS 12.6 by installing llvm@12 using homebrew and using it by setting the environment vars `CC` and `CXX` to point to it.

malishav:
@kaspar030 I was trying use this PR by issuing

`cargo test --no-default-features --features "hacspec-psa"`

i.e. without the `--package edhoc-rs` option, which should in my understanding run tests for all default members of the workspace, and link `edhoc-hacspec` against `edhoc-crypto-psa` backend. But it fails to build because it doesn't seem to find the PSA stubs. Do you understand why?

kaspar030:
> i.e. without the `--package edhoc-rs` option, which should in my understanding run tests for all default members of the workspace, and link `edhoc-hacspec` against `edhoc-crypto-psa` backend. But it fails to build because it doesn't seem to find the PSA stubs. Do you understand why?

nope :( I'll take a look

kaspar030:
I don't get it how features are applied to workspaces. I think what happens is that they get only supplied to the packages that support them. So, building the workspace member "edhoc-rs" with nodefaultfeatures and the hacspec-psa succeeds.
Then cargo goes on to building just hacspec with nodefaultfeatures. hacspec doesn't have "hacspec-psa" as feature, thus fails.

We could reduce the workspace "default members" to just edhoc-rs. That would not build the individual crates, only when selected as dependency by edhoc-rs depending on the features used.

kaspar030:
see https://github.com/openwsn-berkeley/edhoc-rs/pull/32, does that fix this for you?

PR #32, opened 2022-11-14
Reduce workspace default members

kaspar030:
This reduces the workspace default members to lib/ and examples/coap.
That allows build-testing the available crypto backends for std builds.
(Fixes issue pointed out at https://github.com/openwsn-berkeley/edhoc-rs/pull/30#issuecomment-1311925707.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,10 +15,6 @@ members = [
 # tested on the host architecture
 default-members = [
   "lib",
-  "hacspec",
-  "crypto",
-  "crypto/edhoc-crypto-hacspec",
-  "crypto/edhoc-crypto-psa",
   "examples/coap",
 ]
 
--- a/examples/coap/Cargo.toml
+++ b/examples/coap/Cargo.toml
@@ -4,6 +4,11 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-edhoc-rs = { path = "../../lib", features = [ "hacspec-native" ] }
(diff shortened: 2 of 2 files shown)


malishav:
With this PR, in order to run the unit tests in `hacspec/src/lib.rs`, it is in all cases necessary to build-test with

`cargo test --package edhoc-rs --package edhoc-hacspec --no-default-features --features "hacspec-native"`

with a variation on the crypto backend being selected by --features option. Before, it was enough to run `cargo test` on the workspace level. I think I prefer to keep the default as it was before, in order for `cargo test` to invoke these tests by default, and not only when explicitly selected by the crypto backend. Let me know if you disagree or if I am missing osmething

kaspar030:
> Let me know if you disagree or if I am missing osmething

You're right, I only ran a "cargo test" to see if they pass, didn't realize the hacspec tests weren't run.

Maybe we have to re-think the features. IIUC currently they mean:
- hacspec-native => hacspec edhoc, hacspec crypto
- hacspec-psa => hacspec edhoc, psa crypto
- ...

If we split the features (maybe introducing a layer) into edhoc implementation (currently in the branch there's only "hacspec") and crypto implementation, it might work better. I'll give it a try.

malishav:
sounds good!

geonnave:
@malishav should we close this?

malishav:
Closing as obsolete

PR #38, opened 2023-04-03
Add Docs

geonnave:
For now just updating the README, adding a `.gitignore`, and adding a print statement to the example `coapserver.rs`.

@malishav, it looks like libraries in Rust should not have `Cargo.lock` checked into version control, can you confirm that? (if it is the case, I will add it to the `.gitignore` as well).


README.md
```diff
@@ -9,8 +9,29 @@ Supports:
...
+cargo build
+cargo test
+```
 
 ## Example
 
-TODO
+In one terminal, run an example CoAP server:
+```
+./target/debug/coapserver
```
malishav, line 25: Let's use `cargo run --bin` for this

README.md
```diff
@@ -9,8 +9,29 @@ Supports:
 
 ## Installation
 
-TODO
+1. Make sure you have Rust installed (see instructions in the [website](https://www.rust-lang.org/tools/install)).
+
+2. Download, compile, and run the tests:
+```
+git clone git@github.com:geonnave/edhoc-rs.git && cd edhoc-rs
```
malishav, line 16: Let's use the official repo here, `openwsn-berkeley/edhoc-rs`

README.md
```diff
@@ -9,8 +9,29 @@ Supports:
...
 
-TODO
+In one terminal, run an example CoAP server:
+```
+./target/debug/coapserver
+```
+
+In another one, run the client:
+```
+./target/debug/coapclient
```
malishav, line 30: Same comment as above

malishav (commented):
Thanks @geonnave for providing the PR !

malishav:
> @malishav, it looks like libraries in Rust should not have Cargo.lock checked into version control, can you confirm that? (if it is the case, I will add it to the .gitignore as well).

Yes, I think we can add `Cargo.lock` to `.gitignore`

geonnave:
I think the README docs are enough for now, since many things are likely to change, such as the build system parameters.

malishav:
Looks great! Let's just add steps on building and flashing for nrf52840 with and without hardware support for now. That should be pretty useful for a newcomer to the repo.

geonnave:
Good idea, done.

malishav:
Looks great, merging

Issue #42, opened 2023-04-05
Use RNG to generate ephemeral keys


malishav:
This issue depends on #41.

The current protocol implementation in both hacspec and Rust versions uses hardcoded ephemeral keys. This issue aims at using the RNG provided by the crypto API in order to properly generate ephemeral keys for each protocol execution.

See:
- https://github.com/openwsn-berkeley/edhoc-rs/blob/main/hacspec/src/lib.rs#L128 and https://github.com/openwsn-berkeley/edhoc-rs/blob/main/hacspec/src/lib.rs#L316 for hacspec version
- https://github.com/openwsn-berkeley/edhoc-rs/blob/main/lib/src/edhoc.rs#L126 and https://github.com/openwsn-berkeley/edhoc-rs/blob/main/lib/src/edhoc.rs#L308 for Rust version

malishav:
Resolved with #45

Issue #41, opened 2023-04-05
Extend the crypto API to include random number generation


malishav:
This issue aims at extending the crypto API in `/crypto` in order to add support for generating random numbers.

malishav:
Resolved with #45

PR #45, opened 2023-04-06
Add RNG and generate ephemeral keys

geonnave:
Addressing #41 and #42.

This PR adds support for generation of ephemeral private keys, as well as derivation of public keys. It also updates `prepare_message_1` and `prepare_message_2` to accept an ephemeral keypair as parameter.

The implementation differs for each crypto backend:
- hacspec: relies on the [rand](https://crates.io/crates/rand) crate
- psa: uses [psa_crypto::operations::key_management::generate](https://docs.rs/psa-crypto/0.10.0/psa_crypto/operations/key_management/fn.generate.html)
- cryptocell310: uses [CRYS_ECPKI_GenKeyPair](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.sdk5.v12.2.0%2Fgroup__crys__ecpki__kg.html)
- cc2538: just a mock for now


malishav (commented):
I don't understand why you copy the code of `pub fn p256_validate_private_key` from https://github.com/hacspec/hacspec/blob/master/examples/p256/src/p256.rs and not call that function directly? Is it because of the difference in types?

geonnave:
Yes, it is because of the different types.

malishav:
Before invoking the hacspec function, why don't you convert the array to `ByteSeq` like in https://github.com/openwsn-berkeley/edhoc-rs/blob/main/crypto/edhoc-crypto-hacspec/src/lib.rs#L50? This would avoid code duplication from hacspec

geonnave:
You are right, fixed.

crypto/edhoc-crypto-hacspec/src/lib.rs
```diff
@@ -93,3 +95,29 @@ pub fn p256_ecdh(
...
+    private_key
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_p256_generate_private_key() {
+        let random_key = p256_generate_private_key();
```
malishav, line 120: Since `p256_generate_private_key` returns `BytesP256ElemLen`, which is of length `P256_ELEM_LEN`, this test will always pass, right?

geonnave: You are right. I can't think of other tests to add here. Maybe just remove it?

malishav: Same here, agreed

crypto/edhoc-crypto-hacspec/src/lib.rs
```diff
@@ -77,6 +78,7 @@ pub fn aes_ccm_decrypt_tag_8(
 
     (err, p3)
 }
+
```
malishav, line 81: remove newline :)

crypto/edhoc-crypto-hacspec/Cargo.toml
```diff
@@ -14,3 +14,4 @@ hacspec-hkdf = { version = "0.1.0" }
 hacspec-sha256 = { version = "0.1.0" }
 hacspec-aes = { version = "0.1.0" }
 hacspec-aes-ccm = { version = "0.1.0" }
+rand = "0.8.5"
```
malishav, line 17: add newline :)

crypto/edhoc-crypto-hacspec/src/lib.rs
geonnave: This was actually intentional, since there was no blank line between the functions (and having such a blank line between functions seems to be an idiom in the codebase).

By the way, what do you think of using [rustfmt](https://doc.rust-lang.org/book/appendix-04-useful-development-tools.html) ?

malishav: Sounds good! All for `rustfmt`, you can just issue `cargo fmt` command for it to do its job

geonnave: great

examples/edhoc-rs-no_std/src/main.rs
```diff
@@ -13,7 +13,8 @@ use panic_semihosting as _;
 #[cfg(feature = "rtt")]
 use rtt_target::{rprintln as println, rtt_init_print};
 
-use edhoc_rs::{EDHOCError, EdhocInitiator, EdhocResponder, EdhocState};
+// use edhoc_rs::{EDHOCError, EdhocInitiator, EdhocResponder, EdhocState};
```
malishav, line 16: remove dead code

malishav:
Since this is now implemented, could you remove the `G_X, X, G_Y, Y` constants from `consts/src/lib.rs` for both modules?

geonnave:
> Since this is now implemented, could you remove the G_X, X, G_Y, Y constants from consts/src/lib.rs for both modules?

I can remove `Y` and `G_Y`, as I am still using `X` and `G_Y` for tests.

geonnave:
> But why don't you move `X` and `G_X` then as test vectors in the test module? They should definitely not appear any more in `consts`.

Sounds good, will do it.

malishav:
Congrats on getting this ready for merge! Merging now!

Issue #52, opened 2023-04-18
Remove hacspec-specific code from no_std example


geonnave:
It fails when running `cargo build --target="thumbv7em-none-eabihf" --no-default-features --features="rust-psa, rtt" --release`

malishav:
@geonnave Can I close this issue?

PR #54, opened 2023-04-18
fix: make sure the no_std example works on rust and hacspec

geonnave:
Addressing issue #52.

@malishav I was going to just remove the `test_p256_keys` function, but I think it is useful to have it testing key generation here. What do you think?


malishav:
Yes, it's an important test to have. In the absence of a better solution, I think we can merge it.

Issue #50, opened 2023-04-18
Check message length in decode_plaintext_2


malishav:
Add error handling to `decode_plaintext_2` function

geonnave:
Addressed in #58 .

malishav:
Resolved with #58

PR #58, opened 2023-04-21
Check message length in decode_plaintext_2

geonnave:
Addressing #50.

Also, adding some documentation to core functions, and changing parameter names in `compute_prk_3e2m`.


malishav:
Looks good, should I merge?

PR #61, opened 2023-04-25
ci: configure automated releases

geonnave:


geonnave:
To test this, I need to send a PR from the original repo, since it RELEASE_TOKEN is not available for forked repos.

PR #62, opened 2023-04-25
ci: configure automated releases

geonnave:
Automatically create releases (and release notes) from tags.

The only needed manual work is to create and push a git tag. For example:
```bash
git tag REL-0.1.0
git push origin REL-0.1.0
```

The above will generate a new release named `REL-0.1.0`.

**Note**: the tag must be pushed from the original repo, and not from a fork, as forks have restricted permissions.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,11 @@
+changelog:
+  categories:
+    - title: New features
+      labels:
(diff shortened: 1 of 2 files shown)


geonnave:
@malishav PR is now ready.

PR #63, opened 2023-04-26
Get rid of compiler warnings

geonnave:
Get rid of warnings, most being about (1) values assigned but not read or (2) unecessary use of `mut`.

For now, just adjusted the hacspec version. If ok, I will adjust the rust version as well.

### DIFF
--- a/hacspec/src/lib.rs
+++ b/hacspec/src/lib.rs
@@ -24,11 +24,8 @@ pub fn edhoc_exporter(
         _th_3,
     ) = state;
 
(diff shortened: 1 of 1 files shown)


malishav:
Note that most of the occurrences that caused warnings were there to be compatible with hacspec v1. Before merging, let's make sure  it passes hacspec v2 checks.

geonnave:
The code here is very old, if we address this it should be in a new PR. Closing.

PR #69, opened 2023-05-03
Variable size messages

geonnave:
Addressing #66.

### DIFF
--- a/hacspec/src/lib.rs
+++ b/hacspec/src/lib.rs
@@ -58,41 +58,48 @@ pub fn r_process_message_1(
     if current_state == EDHOCState::Start {
         // Step 1: decode message_1
         // g_x will be saved to the state
-        let (method, supported_suites, g_x, c_i) = parse_message_1(message_1);
+        let res: Result<(U8, BytesSupportedSuites, BytesP256ElemLen, U8), EDHOCError> =
+            parse_message_1(message_1);
 
-        // verify that the method is supported
-        if method.declassify() == EDHOC_METHOD {
-            // Step 2: verify that the selected cipher suite is supported
-            if supported_suites[0u8].declassify() == EDHOC_SUPPORTED_SUITES[0u8].declassify() {
-                // Step 3: If EAD is present make it available to the application
-                // TODO we do not support EAD for now
+        if res.is_ok() {
+            let (method, supported_suites, g_x, c_i) = res.unwrap();
 
-                // hash message_1 and save the hash to the state to avoid saving the whole message
-                h_message_1 = sha256_digest(
-                    &BytesMaxBuffer::from_slice(&message_1.content, 0, message_1.len),
-                    message_1.len,
-                );
+            // verify that the method is supported
+            if method.declassify() == EDHOC_METHOD {
+                // Step 2: verify that the selected cipher suite is supported
+                if supported_suites[0u8].declassify() == EDHOC_SUPPORTED_SUITES[0u8].declassify() {
+                    // Step 3: If EAD is present make it available to the application
+                    // TODO we do not support EAD for now
 
-                error = EDHOCError::Success;
-                current_state = EDHOCState::ProcessedMessage1;
+                    // hash message_1 and save the hash to the state to avoid saving the whole message
+                    h_message_1 = sha256_digest(
+                        &BytesMaxBuffer::from_slice(&message_1.content, 0, message_1.len),
+                        message_1.len,
+                    );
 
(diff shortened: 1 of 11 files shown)


geonnave:
The Rust version is now complete -- messages can have variable size.

Next: add it to the hacspec version.

geonnave:
Hacspec version works, last thing missing is handling several cipher suites in hacspec version.

hacspec/src/lib.rs
```diff
@@ -607,59 +609,122 @@ pub fn construct_state(
...
+) -> Result<(U8, BytesSupportedSuites, BytesP256ElemLen, U8), EDHOCError> {
+    let mut error: EDHOCError = EDHOCError::UnknownError;
+    let method = rcvd_message_1.content[0];
+    let mut g_x: BytesP256ElemLen = BytesP256ElemLen::new();
+    let mut proposed_suites: BytesSuites = BytesSuites::from_public_array([0xff; SUITES_LEN]);
+    let mut proposed_suites_len: usize = 0;
+    let mut selected_suites: BytesSupportedSuites = BytesSupportedSuites::new();
+    let mut c_i = U8(0);
+
+    let result = match (rcvd_message_1.content[1] as U8).declassify() {
```
malishav, line 622: Please comment that you are matching based on first byte of `SUITES_I`, which can be either an `int` or an array

hacspec/src/lib.rs
```diff
@@ -607,59 +609,122 @@ pub fn construct_state(
...
+            proposed_suites[0] = rcvd_message_1.content[2];
+            Ok(2)
+        }
+        0x80..=0x97 => {
+            let proposed_suites_len: U8 = rcvd_message_1.content[1] - U8(0x80);
+            let proposed_suites_len: usize = proposed_suites_len.declassify().into();
+            for i in 0..proposed_suites_len {
+                proposed_suites[i] = rcvd_message_1.content[2 + i];
+            }
+            Ok(proposed_suites_len + 1)
```
malishav, line 637: why +1 here?

hacspec/src/lib.rs
```diff
@@ -607,59 +609,122 @@ pub fn construct_state(
...
+        }
+        _ => Err(EDHOCError::ParsingError),
+    };
+
+    if result.is_ok() {
+        proposed_suites_len = result.unwrap();
+        let proposed_suites = proposed_suites.to_public_array();
+        let mut selected_suites_len: usize = 0;
+        for i in 0..EDHOC_SUPPORTED_SUITES.len() {
+            if proposed_suites.contains(&EDHOC_SUPPORTED_SUITES[i].declassify()) {
```
malishav, line 647: This code does not belong in a parsing routine. Parsing routine should just return what was proposed. The invoking code should check whether the proposed suite is indeed supported and act accordingly.

malishav (commented):
Thanks for providing this PR, it is massive! I would be more confident in first merging the variable-length message code, without the code on cipher suites. As you will see inline, most of my comments are on that part of the code. Would it be possible to separate the two?

geonnave:
Thanks for the review, I have reverted the changes about cipher suite parsing. I will consider your comments and add them to the other PR that only deals with that part.

hacspec/src/lib.rs
```diff
@@ -831,56 +835,54 @@ fn compute_k_3_iv_3(
...
-            error = p3.err().expect("error handling error");
-        }
+    if p3.is_ok() {
+        error = EDHOCError::Success;
+        let p3 = p3.unwrap();
+        plaintext_3.content = plaintext_3.content.update_slice(0, &p3.content, 0, p3.len);
+        plaintext_3.len = p3.len;
     } else {
-        error = EDHOCError::ParsingError;
+        error = p3.err().expect("error handling error");
```
malishav, line 885: we should replace this and similar lines to use `unwrap_err`

hacspec/src/lib.rs
```diff
@@ -1092,8 +1091,12 @@ mod tests {
     // manually modified test vector to include a single supported cipher suite
     const MESSAGE_1_TV: &str =
         "030258208af6f430ebe18d34184017a9a11bf511c8dff8f834730b96c1b7c8dbca2fc3b637";
+    // manually modified test vector to include two supported cipher suites (02 and 03), encoded as array
```
malishav, line 1094: likely a remnant comment?

geonnave:
All comments have been addressed.

malishav (approved):

Issue #65, opened 2023-05-02
Support for cipher suite negotiation


malishav:
The code base currently expects a single cipher suite and will fail if anything else is received (e.g. multiple supported suites). This issue aims to add support for parsing multiple supported suites although only one is supported.

malishav:
Resolved with #71 , closing

PR #71, opened 2023-05-10
Support parsing multiple cipher suites

geonnave:
Addressing #65.

### DIFF
--- a/hacspec/src/lib.rs
+++ b/hacspec/src/lib.rs
@@ -58,41 +58,48 @@ pub fn r_process_message_1(
     if current_state == EDHOCState::Start {
         // Step 1: decode message_1
         // g_x will be saved to the state
-        let (method, supported_suites, g_x, c_i) = parse_message_1(message_1);
+        let res = parse_message_1(message_1);
 
-        // verify that the method is supported
-        if method.declassify() == EDHOC_METHOD {
-            // Step 2: verify that the selected cipher suite is supported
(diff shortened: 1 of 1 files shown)


hacspec/src/lib.rs
```diff
@@ -601,17 +608,112 @@ pub fn construct_state(
     )
 }
 
+fn parse_cipher_suites(
```
malishav, line 611: Since the function is specific to parsing `SUITES_I` in `message_1` consider renaming it accordingly. e.g. `parse_suites_i`

malishav (commented):
Thanks for the PR! Couple of comments inline. This made me think that we will at some point need to add the protocol logic for sending back the error, as per Section 6.3 of draft-ietf-lake-edhoc-19. This does not need to be in this PR, but let's track it in the issues.

geonnave:
Comments addressed, Rust version pushed.

malishav (commented):
One last comment before merging which I just realized: Could you add the unmodified `message_1` test vector from draft-ietf-lake-traces-05. Previously, I had to modify it manually to have a single suite in the message. Now, it should just work as-is. The test vector for message_1 is:

`03 82 06 02 58 20 8a f6 f4 30 eb e1 8d 34 18 40 17 a9 a1 1b f5 11 c8
df f8 f8 34 73 0b 96 c1 b7 c8 db ca 2f c3 b6 37`

geonnave:
Test vector updated.

Also had to do some changes in `encode_message_1` and found / fixed a bug in `parse_suites_i`.

PR #73, opened 2023-05-22
Add ead processing mechanics

geonnave:
Work in progress.

- Introduces a new feature called `ead-zeroconf`, which enables the zero conf implementation
- For now at least, it also needs a `ead-none` feature, which is used when no EAD is to be used

How to test:

```
RUST_BACKTRACE=1 cargo test --no-default-features --package edhoc-rs --package edhoc-crypto --package edhoc-hacspec --package edhoc-consts --features="hacspec-psa, ead-zeroconf" --no-fail-fast test
```

### DIFF
--- a/hacspec/src/lib.rs
+++ b/hacspec/src/lib.rs
@@ -23,7 +23,7 @@ pub fn edhoc_exporter(
         _h_message_1,
         _th_3,
         _ead_init_handler,
-        _ead_rcvr_handler,
+        ead_resp_handler,
     ) = state;
 
     let mut output = BytesMaxBuffer::new();
@@ -74,7 +74,10 @@ pub fn r_process_message_1(
                 {
                     // Step 3: If EAD is present make it available to the application
                     if let Some(mut ead_handler) = ead_resp_handler {
-                        ead_handler.state = (ead_handler.process_ead1_cb)(message_1.to_public_buffer(), ead_handler.state);
+                        ead_handler.state = (ead_handler.process_ead1_cb)(
+                            message_1.to_public_buffer(),
+                            ead_handler.state,
+                        );
                         ead_resp_handler = Some(ead_handler);
                     }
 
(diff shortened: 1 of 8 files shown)


geonnave:
There are a few things still to improve, but could already receive a review.

Main things to improve:
- [ ] code duplication in EAD processing
- [ ] use smaller buffer sizes (using `const generics` would be ideal, see if it can be used with `hacspec`)

hacspec/src/lib.rs
```diff
@@ -232,72 +278,115 @@ pub fn r_process_message_3(
...
+                        }
+                    } else {
+                        error = EDHOCError::UnknownPeer;
+                    }
+                } else {
+                    error = EDHOCError::EADHandlingFailed;
                 }
             } else {
-                error = EDHOCError::UnknownPeer;
+                error = decoded_p3_res.unwrap_err();
```
geonnave, line 389: Undo this change.

malishav (commented):
I have three high-level comments on this PR:

1) We should split the PR into two PRs. One for parsing and encoding routines on EAD items, which seems almost ready to merge, and another one for the mechanics on EAD handlers and EAD processing in top-level processing functions.

2) From parsing routines, could you return a tuple (int, ByfferEadN) where integer corresponds to the EAD label. Based on this label, we should be able to know whether we support the processing of the current EAD item. Now, you do partial parsing in the parsing routines and partial in the top-level e.g. process_message_3 routine (Line 299 in `hacspec/src/lib.rs`. We should avoid that.

3) EAD structs that you currently define in crate `consts` should really live in a separate crate, without the need to define `ead-none`  and `ead-zeroconf` features on both `hacspec` crate and `consts` crate. I would propose creating a directory/crate `ead`, just like it is done for `crypto` and `consts` and in that crate have a top-level `lib.rs` which simply multiplexes between `ead-none` and `ead-zeroconf` based on the defined feature *on that `ead` crate*. Then, each EAD item handler should be a separate file within that crate: `ead-none` and `ead-zeroconf` at the moment. By default, that crate compiles for `ead-none`, and none-handlers are imported. If that crate is compiled with `ead-zeroconf` feature, then different handlers are imported.

PR #74, opened 2023-05-26
Parse ead items

geonnave:
Parse and encode EAD items in messages 1, 2, and 3.

Implemented in hacspec version, ready for review. After the review I will bring the changes to the Rust version.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -110,6 +110,14 @@ mod common {
 mod rust {
     use super::common::*;
 
+    #[derive(Debug)]
+    pub struct EADItem {
+        pub label: u8,
+        pub is_critical: bool,
+        // TODO[ead]: have adjustable (smaller) length for this buffer
+        pub value: Option<EdhocMessageBuffer>,
+    }
+
     pub type U8 = u8;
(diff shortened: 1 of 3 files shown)


geonnave:
I added parsing of the ead label and value (for now only in parse_message_1), with the caveat that it assumes only one EAD item.

consts/src/lib.rs
```diff
@@ -210,7 +210,22 @@ mod hacspec {
         }
     }
 
-    array!(BytesEad2, 0, U8);
+    #[derive(Debug)]
+    pub struct BufferEAD {
```
malishav, line 214: Let's call it `EADItem` to be consistent with the spec

malishav:
Great! Make sure just that you do not duplicate the parsing code, since this will be the same for other EAD items.

consts/src/lib.rs
```diff
@@ -210,7 +210,22 @@ mod hacspec {
...
+        pub label: i8,
+        pub value: EdhocMessageBufferHacspec,
```
malishav, line 216: Should `value` be an `Option`? It is optionally specified in the spec

geonnave:
All should be ready now.

PR #75, opened 2023-05-26
EAD mechanics using separate crate and enabled via features

geonnave:
Addressing #67.

This PR introduces a new crate, `edhoc-ead`, which has two subcrates: `edhoc-ead-none` and `edhoc-ead-zeroconf`. The former is a default placeholder that provides mocked functions to satisfy Rust's compiler, while the latter provides a skeleton implementation for the [lake-authz draft](https://ericssonresearch.github.io/ace-ake-authz/draft-selander-lake-authz.html). This structure is just like what is used with `edhoc-crypto` to enable multiple crypto backends.

Some remarks:
- achieved goal of keeping stack-only allocations
- in the current approach, only one EAD handler implementation is supported, e.g., `edhoc-rs` can be compiled with one of: `ead-none`, `ead-zeroconf`, `ead-foobar`, etc.
- currently, the EAD state is global and mutable, and not in a thread safe way -- should not be a problem for now since we only have single-threaded code, but could be improved.

### DIFF
--- a/ead/edhoc-ead-zeroconf/Cargo.toml
+++ b/ead/edhoc-ead-zeroconf/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "edhoc-ead-zeroconf"
+version = "0.1.0"
+edition = "2021"
+authors = ["Geovane Fedrecheski <geonnave@gmail.com>"]
+license = "BSD"
+description = "EDHOC EAD zeroconf (draf-lake-authz)"
+
+[dependencies]
+edhoc-consts = { path = "../../consts" }
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -29,6 +29,7 @@ jobs:
       matrix:
         edhoc_lib: [hacspec, rust]
         crypto: [hacspec, psa]
+        ead: [ead-none, ead-zeroconf]
         exclude:
           - edhoc_lib: rust
             crypto: hacspec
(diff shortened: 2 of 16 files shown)


geonnave:
Finally working, I think the main goal of this PR has been reached.

Since some features changed, here are some of the updated commands to build/run/test (might add them to the README).

#### Test without ead (ead-none, the default):
```bash
cargo test --package edhoc-rs --package edhoc-crypto --package edhoc-hacspec --package edhoc-ead  --features="hacspec-psa" --no-fail-fast
````

#### Test with ead-zeroconf:
```bash
cargo test --package edhoc-rs --package edhoc-crypto --package edhoc-ead --no-default-features  --features="rust-psa, ead-zeroconf" --no-fail-fast
````
Note that `--no-default-features` is needed.

#### Flash to device with zeroconf:
```bash
cargo embed --target="thumbv7em-none-eabihf" --no-default-features --features="rust-cryptocell310, ead-zeroconf, rtt" --release
````

File: ead/edhoc-ead-zeroconf/Cargo.toml [hunk 1]
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "edhoc-ead-zeroconf"
```
geonnave, line 2: Maybe rename to `edhoc-ead-zerotouch` ?

malishav (approved):
This seems ready to merge

geonnave:
PR is merged ([see here](https://github.com/openwsn-berkeley/edhoc-rs/commits/main)), although GitHub UI seems to disagree :upside_down_face:. Let's wait a bit, might be a temporary bug on GitHub side.

geonnave:
The GitHub UI insists it is not merged yet. Should I just close it?

malishav:
Yes, please do

PR #84, opened 2023-06-20
Enable integration with C projects

geonnave:
The approach is as follows:

- modify `Cargo.toml` to also generate a static library (`libedhoc_rs.a`)
- build a FFI-friendly wrapper, so that the relevant functions can be called with C types
- annotate functions with `#[no_mangle] pub extern "C"` and structs with `#[repr(C)]`
- configure [cbindgen](https://github.com/mozilla/cbindgen) so that it automatically generates C headers for the FFI-friendly wrapper
  - ~~it runs automatically on every build (cbindgen is configured in `build.rs`)~~

~~Now every build will also build the static library and the appropriate headers.~~

A script is provided to facilitate the generation of static library and the headers. Example usage:
```bash
./build_for_c.sh rust-cryptocell310
```

Here is the basic usage form C:

```c
static const uint8_t ID_CRED_I[] = "a104412b";
static const uint8_t ID_CRED_R[] = "a104410a";
static const uint8_t CRED_I[] = "A2027734322D35302D33312D46462D45462D33372D33322D333908A101A5010202412B2001215820AC75E9ECE3E50BFC8ED60399889522405C47BF16DF96660A41298CB4307F7EB62258206E5DE611388A4B8A8211334AC7D37ECB52A387D257E6DB3C2A93DF21FF3AFFC8";
static const uint8_t G_I[] = "ac75e9ece3e50bfc8ed60399889522405c47bf16df96660a41298cb4307f7eb6";
static const uint8_t CRED_R[] = "A2026008A101A5010202410A2001215820BBC34960526EA4D32E940CAD2A234148DDC21791A12AFBCBAC93622046DD44F02258204519E257236B2A0CE2023F0931F1F386CA7AFDA64FCDE0108C224C51EABF6072";
static const uint8_t R[] = "72cc4761dbd4c78f758931aa589d348d1ef874a7e303ede2f140dcf3e6aa4aac";
static const uint8_t I[] = "fb13adeb6518cee5f88417660841142e830a81fe334380a953406a1305e8706b";
static const uint8_t G_R[] = "bbc34960526ea4d32e940cad2a234148ddc21791a12afbcbac93622046dd44f0";

RustEdhocInitiatorC initiator = initiator_new(I, 32*2, G_R, 32*2, ID_CRED_I, 4*2, CRED_I, 107*2, ID_CRED_R, 4*2, CRED_R, 84*2);
RustEdhocResponderC responder = responder_new(R, 32*2, G_I, 32*2, ID_CRED_I, 4*2, CRED_I, 107*2, ID_CRED_R, 4*2, CRED_R, 84*2);

EdhocMessageBuffer message_1;
initiator_prepare_message_1(&initiator, &message_1);
responder_process_message_1(&responder, &message_1);
EdhocMessageBuffer message_2;
uint8_t c_r_sent;
responder_prepare_message_2(&responder, &message_2, &c_r_sent);
uint8_t c_r_received;
initiator_process_message_2(&initiator, &message_2, &c_r_received);
EdhocMessageBuffer message_3;
uint8_t prk_out_initiator[SHA256_DIGEST_LEN];
initiator_prepare_message_3(&initiator, &message_3, prk_out_initiator);
uint8_t prk_out_responder[SHA256_DIGEST_LEN];
responder_process_message_3(&responder, &message_3, prk_out_responder);
```


geonnave:
Some demo code is available in the `DotBot-firmware` repository, in the [integrate-edhoc-rs](https://github.com/geonnave/DotBot-firmware/tree/integrate-edhoc-rs) branch.

include/edhoc_rs.h
```diff
@@ -0,0 +1,93 @@
+/*
```
malishav, line 1: I don't understand why is this file included in the repo when it seems to be generated on the fly by the `build_for_c.sh` script?

include/edhoc_consts.h
```diff
@@ -0,0 +1,97 @@
+/*
```
malishav, line 1: Same comment as for `edhoc_rs.h`. Why do you include this in the repo when it should be generated on the fly?

examples/c-wrapper-riot/main.c
```diff
@@ -0,0 +1,62 @@
...
+    EdhocMessageBuffer message_3;
+    uint8_t prk_out_initiator[SHA256_DIGEST_LEN];
+    initiator_prepare_message_3(&initiator, &message_3, &prk_out_initiator);
+    uint8_t prk_out_responder[SHA256_DIGEST_LEN];
+    responder_process_message_3(&responder, &message_3, &prk_out_responder);
+
+    printf("\nprk_out_initiator: \n");
+    od_hex_dump(prk_out_initiator, SHA256_DIGEST_LEN, OD_WIDTH_DEFAULT);
+    printf("\nprk_out_responder: \n");
+    od_hex_dump(prk_out_responder, SHA256_DIGEST_LEN, OD_WIDTH_DEFAULT);
```
malishav, line 57: Wouldn't it make sense here to do a `memcmp` of `prk_out_responder` and `prk_out_initiator` and return 1 if they don't match, 0 if all ok?

lib/Cargo.toml
```diff
@@ -15,16 +15,24 @@ hacspec-lib = { version = "0.1.0-beta.1", default-features = false, optional = t
...
+[build-dependencies]
+cbindgen = "0.24.5"
 
 [features]
 default = [ "edhoc-ead/ead-none" ]
 hacspec-hacspec = ["hacspec-lib/std", "edhoc-hacspec", "edhoc-crypto/hacspec", "edhoc-consts/hacspec" ]
 hacspec-cc2538 = ["hacspec-lib/alloc", "edhoc-hacspec", "edhoc-crypto/cc2538", "edhoc-consts/hacspec" ] # FIXME: stubs
-hacspec-psa = ["hacspec-lib/alloc", "edhoc-hacspec", "edhoc-crypto/psa", "edhoc-consts/hacspec" ]
-hacspec-psa-baremetal = ["hacspec-psa", "edhoc-crypto/psa-baremetal", "edhoc-consts/hacspec" ]
+hacspec-psa = ["hacspec-lib/alloc", "edhoc-hacspec", "edhoc-crypto/psa", "edhoc-consts/hacspec", "embedded-alloc" ]
```
malishav, line 28: I am missing why is it needed to add `embedded-alloc` as a dependency in this file?

malishav (commented):
Thanks for the PR, it seems almost ready to fly. Couple of questions inline, some are mainly for my understanding.

include/edhoc_rs.h
geonnave: Yes, I just wanted to put the header files _somewhere_. Maybe:
- move the generated include folder to `./target/include` and
- add to the releases as a generated artefact, just like the fstar files?

include/edhoc_consts.h
geonnave: Same as above.

examples/c-wrapper-riot/Makefile
```diff
@@ -0,0 +1,27 @@
...
+endif
+
+# This is actually only needed in the RUST_CRYPTOCELL310 configuration
+CFLAGS += -DTHREAD_STACKSIZE_DEFAULT=16384 -DISR_STACKSIZE=16384
+
+USEMODULE += od
+USEMODULE += ps
+
+# This has to be the absolute path to the RIOT base directory:
+RIOTBASE ?= $(CURDIR)/../../../RIOT
```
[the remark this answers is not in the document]
geonnave: Yes I am making that assumption. Added an explanation to the README.

include/edhoc_rs.h
geonnave: Moved to `target/include`.

examples/c-wrapper-riot/main.c
geonnave: Agree, done.

lib/Cargo.toml
geonnave: This is a remnant of when I was trying to have `staticlib` always in Cargo.toml (as opposed to only addding it when I want to build the static library, using `build_for_c.sh`).

It is not needed anymore, good catch.

lib/Cargo.toml
```diff
@@ -15,6 +15,10 @@ hacspec-lib = { version = "0.1.0-beta.1", default-features = false, optional = t
 edhoc-crypto = { path = "../crypto", default-features = false }
 edhoc-consts = { path = "../consts", default-features = false }
 edhoc-ead = { path = "../ead", default-features = false }
+panic-semihosting = { version = "0.6.0", features = ["exit"], optional = true }
```
malishav, line 18: Why do you need `panic-semihosting` here? Can it also be removed?

geonnave: When I try to remove it, the `staticlib` build complains that it needs a panic handler. One alternative would be just having an empty panic handler with `loop {}`, but I don't know if it can be replace by a better one later in the application, so I decided to put this in the lib.

But I think I can check about whether an application-defined panic holder would override an empty one.

PR #85, opened 2023-06-20
Generate dynamic connection identifiers

geonnave:


malishav:
Should we merge this? It looks good to me.

malishav:
One small nit: I think constant `C_R` is still present in `consts.rs`

geonnave:
Thanks for the review and sorry for the delay. Deleted `C_I` and `C_R` from `consts/src/lib.rs`. Now for my part I think it is ready for merge.

PR #95, opened 2023-09-13
Improve parsing robustness

geonnave:


hacspec/src/lib.rs
```diff
@@ -855,46 +884,62 @@ fn parse_message_1(
...
     let mut suites_i_len: usize = 0;
     let mut raw_suites_len: usize = 0;
     let mut c_i = U8(0);
     let mut ead_1 = None::<EADItemHacspec>;
 
-    let method = rcvd_message_1.content[0];
-
-    let res_suites = parse_suites_i(rcvd_message_1);
+    // check surplus array encoding
+    if !is_cbor_array(rcvd_message_1.content[0]) {
```
malishav, line 896: I don't understand the logic here. You are returning `ParsingError` if it is a CBOR array. What if it is a map? Or something else?

hacspec/src/lib.rs
```diff
@@ -1668,6 +1747,27 @@ mod tests {
         assert_eq!(c_i.declassify(), c_i_tv.declassify());
     }
 
+    #[test]
+    fn test_parse_message_1_invalid_traces() {
```
malishav, line 1751: All these tests should check the type of the error, i.e. whether it is `ParsingError`, to be explicit.

malishav (commented):
Thanks for working on this. Some discussion comments inline.

hacspec/src/lib.rs
geonnave: Good question. Re-reading the traces document, this should validate that the message is a CBOR sequence, so I think I will instead check for the type of each element.

hacspec/src/lib.rs
```diff
@@ -981,14 +1026,34 @@ fn encode_message_1(
...
     // FIXME decode negative integers as well
-    let g_y = BytesP256ElemLen::from_slice(&rcvd_message_2.content, 2, P256_ELEM_LEN);
-    let ciphertext_2_len = rcvd_message_2.len - P256_ELEM_LEN - 2; // len - gy_len - 2
-    let ciphertext_2 =
-        BufferCiphertext2::from_slice(&rcvd_message_2.content, 2 + P256_ELEM_LEN, ciphertext_2_len);
+    let mut g_y: BytesP256ElemLen = BytesP256ElemLen::new();
+    let mut ciphertext_2: BufferCiphertext2 = BufferCiphertext2::new();
+
+    // ensure the whole message is a single CBOR sequence
+    if u8::from(rcvd_message_2.content[0]) == CBOR_BYTE_STRING
```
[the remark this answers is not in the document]
geonnave: Yes, ajdusted.

hacspec/src/lib.rs
geonnave: Agreed, fixed.

hacspec/src/lib.rs
```diff
@@ -855,46 +884,62 @@ fn parse_message_1(
...
-            // NOTE: since the current implementation only supports one EAD handler,
-            // we assume only one EAD item
-            let ead_res = parse_ead(rcvd_message_1, 4 + raw_suites_len + P256_ELEM_LEN);
-            if ead_res.is_ok() {
-                ead_1 = ead_res.unwrap();
-                error = EDHOCError::Success;
+                // check surplus bstr encoding
+                c_i = rcvd_message_1.content[3 + raw_suites_len + P256_ELEM_LEN];
+                let c_i_lookahead = rcvd_message_1.content[4 + raw_suites_len + P256_ELEM_LEN];
+                if !should_encoded_conn_id_be_minimal(c_i, c_i_lookahead) {
```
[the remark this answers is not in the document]
geonnave: We do, but as per Section [3.3.2.](https://www.ietf.org/archive/id/draft-ietf-lake-edhoc-22.html#section-3.3.2) there are two ways of representing the CID: byte-string or small-integer (-24 up to 23).
If a CID is a small-integer, it must be encoded as integer.
Here I check if the message contains a byte-string of size 1 whose single value is -24 up to 23.
We never generate that kind of CID, but other implementations could.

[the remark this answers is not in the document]
malishav: I see. But your check is really tailored to the invalid test vector we have at hand. It should be general in the sense that we reject anything that is not supported by the application. We still do not support bstr-encoded CID if I am not mistaken, so you could only check whether the current byte is in the range -24 to 23. Anything else is a parsing error. What do you think?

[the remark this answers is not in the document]
geonnave: Yes, that makes sense. We indeed do not support bstr-encoded CID. Adjusted.

hacspec/src/lib.rs
```diff
@@ -755,31 +760,36 @@ fn parse_suites_i(
...
-                    } else if cs_id == 0x18 {
-                        // CBOR unsigned integer (one-byte uint8_t follows)
-                        raw_suites_len += 1; // account for the 0x18 tag byte
-                        suites_i[j] = rcvd_message_1.content[raw_suites_len];
-                        suites_i_len += 1;
-                    } else {
-                        error = EDHOCError::ParsingError;
-                        error_occurred = true;
+        // check surplus array encoding of ciphersuite
+        if suites_len > 1 {
```
malishav, line 764: Any particular reason why the checks on lines 764 and 766 are not grouped?

malishav: What does "surplus" array encoding mean?

hacspec/src/lib.rs
```diff
@@ -855,46 +877,61 @@ fn parse_message_1(
...
+            (suites_i, suites_i_len, raw_suites_len) = res_suites.unwrap();
 
-        g_x = BytesP256ElemLen::from_slice(
-            &rcvd_message_1.content,
-            3 + raw_suites_len,
-            P256_ELEM_LEN,
-        );
-
-        c_i = rcvd_message_1.content[3 + raw_suites_len + P256_ELEM_LEN];
+            let g_x_type = rcvd_message_1.content[1 + raw_suites_len].declassify();
```
malishav, line 897: You could use the similar check here as for `uint8`

hacspec/src/lib.rs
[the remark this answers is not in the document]
malishav: I don't see how this was fixed, you are still using the "lookahead" value. Is that needed?

malishav (commented):
Some further comments. To be consistent with message_2 parsing, could you also add a similar check like in `parse_message_2` to `decrypt_message_3` in order to verify that the whole message is a single CBOR element?

hacspec/src/lib.rs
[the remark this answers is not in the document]
geonnave: It is still present in the Rust version, but I only applied the corrections in the hacspec version since that was where you commented first. I will change the hacspec one then modify the rust version.

hacspec/src/lib.rs
geonnave: It means that the encoding as array uses extra bytes, since it could have been only an integer.

hacspec/src/lib.rs
geonnave: Yes I will do it on all other parsing and decoding routines.

hacspec/src/lib.rs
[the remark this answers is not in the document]
malishav: oops, sounds good!

PR #100, opened 2023-09-29
Privatize EADItem details, increase label range to i16

chrysn:
While it has practical uses (I'd like to utilize the library with https://datatracker.ietf.org/doc/draft-amsuess-core-edhoc-grease/, although that'll need the u16 or practically i32 points), its main purpose is to check whether my understanding of the code is good enough to pass CI.

I'd still ask for reviews towards merging because
* having fewer pub items makes for a better evolvable library, and
* I expect that EAD extensions that are somewhat larger will stay clear of the u8 range to leave it for every-byte-counts applications.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -88,7 +88,7 @@ mod common {
 
     #[derive(Debug)]
     pub struct EADItem {
-        pub(crate) label: u8,
+        pub(crate) label: u16,
         pub(crate) is_critical: bool,
         // TODO[ead]: have adjustable (smaller) length for this buffer
         pub(crate) value: Option<EdhocMessageBuffer>,
@@ -100,9 +100,6 @@ mod common {
     pub enum EADNewError {
         /// The value given exceeds the size that is statically allocated per EAD item.
         SizeExceeded,
-        /// The label value given at construction exceeds the range of EAD labels the library can
-        /// handle.
-        InexpressibleLabel,
         /// The label indicates use as padding, but non-zero values were encountered.
         InvalidPadding,
     }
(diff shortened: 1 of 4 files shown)


consts/src/lib.rs
```diff
@@ -88,30 +88,71 @@ mod common {
...
-        fn new() -> Self {
-            EADItem {
-                label: 0,
-                is_critical: false,
-                value: None,
+        #[inline(always)] // Assist const propagation that removes error states
+        fn new(label: u16, is_critical: bool, value: Option<&[u8]>) -> Result<Self, EADNewError> {
+            let mut label: i16 = label
+                .try_into()
+                .map_err(|_| EADNewError::InexpressibleLabel)?;
```
geonnave, line 126: I think `?` (and early returns in general?) is currently not supported by `cargo hax into fstar`, @W95Psp could you please confirm?

W95Psp: No, that should be fine, `?` and early returns are supported (they simply produce a bit heavyweight F*)

geonnave: Great, thanks for the clarification!

consts/src/lib.rs
```diff
@@ -88,30 +88,71 @@ mod common {
 
     #[derive(Debug)]
     pub struct EADItem {
-        pub label: u8,
-        pub is_critical: bool,
+        pub(crate) label: i16,
```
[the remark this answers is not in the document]
geonnave: Sounds good!

chrysn:
Replaced by https://github.com/openwsn-berkeley/lakers/pull/333

PR #102, opened 2023-09-29
Update dependencies that broke CI

chrysn:
With https://github.com/thvdveld/cc2538-hal/pull/2 closed, cc2538-hal now has a published 0.2.0 that contains the crypto code we previously had a git branch dependency on.

This should fix the build failures that started when the PR was closed and the branch removed.

The -pac crate is updated in lockstep with -hal's dependency to avoid duplicates; the `.take().expect(...)` idiom is exchanged with an unsafe `.steal()` following the examples of cc2538-hal.

Contributes-To: https://github.com/openwsn-berkeley/edhoc-rs/pull/98

CC'ing @thvdveld who is probably the best person to review this.

### DIFF
--- a/crypto/edhoc-crypto-cc2538/Cargo.toml
+++ b/crypto/edhoc-crypto-cc2538/Cargo.toml
@@ -13,8 +13,8 @@ cortex-m = { version = "0.7.4" }
 cortex-m-rt = { version = "0.7.1" }
 cortex-m-semihosting = { version = "0.5.0" }
 panic-semihosting = { version = "0.6.0", features = ["exit"] }
-cc2538-pac = { version = "0.1.0", features = ["rt"] }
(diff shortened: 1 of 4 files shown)


chrysn:
I've updated the PR to also work around a semver incompatible update to the coap crate in https://github.com/Covertness/coap-rs/issues/87, which hit crates.io about 4h ago. Keeping both in one PR as none of them alone has a chance of clearing CI.

chrysn:
One more fix: covertness fixed the coap-rs issue, so to avoid referencing a now yanked version, the forcepushed version now uses the newest CoAP release right away.

malishav:
Let me know if this is ready for merge?

chrysn:
Ready if CI says so.

malishav:
Thanks for providing this PR!

Issue #99, opened 2023-09-28
Role struct constructors


chrysn:
Attempting to use edhoc-rs in a demo application, I found a few properties of the role constructors (eg. `EdhocInitiator`) odd:

* <del>The g_i / g_r is fed in explicitly -- from an application point of view I'd expect the library to either use the platform's random number generator (or, when that is not possible, to provide a random number callback that'll pull as many bytes as needed).</del> [edit: see <https://github.com/openwsn-berkeley/edhoc-rs/issues/99#issuecomment-1740520737>, these are the long-term keys]
* The constructors expect to be fed both sides' credentials -- the responder needs to know in advance which id_cred_i and cred_i the initiator will send, whereas I'd expect EDHOC operations to work with arbitrary initiators. It'd still be up to the application to provide the right CRED_R for an ID_CRED_R, but that could be a map in the simplest case (esp. when KIDs are used), or network based in more complex ones.
* <del>All those binary arguments are passed in hex encoded strings -- a conversion I'd prefer never to do on an embedded system.</del> [edit:fixed in <https://github.com/openwsn-berkeley/edhoc-rs/pull/101>]

I think I could provide patches for most of those, but these might be here for a reason. Could you explain why those are as they are, and whether those are boundary conditions set up by hacspec, or maybe just remnants of earlier development phases? If hacspec tightens things down here, do you have a roadmap for how to provide additional interfaces for constrained devices when hacspec limitations do not apply?

malishav:
Thanks for the feedback. See below.

> The g_i / g_r is fed in explicitly -- from an application point of view I'd expect the library to either use the platform's random number generator (or, when that is not possible, to provide a random number callback that'll pull as many bytes as needed).

Note that these are not the ephemeral keys. These are the public keys present in the corresponding CRED_I/CRED_R. For the moment, we do not parse the credentials for the sake of simplicity but rather use these values in the corresponding ECDH calculations.

> The constructors expect to be fed both sides' credentials -- the responder needs to know in advance which id_cred_i and cred_i the initiator will send, whereas I'd expect EDHOC operations to work with arbitrary initiators. It'd still be up to the application to provide the right CRED_R for an ID_CRED_R, but that could be a map in the simplest case (esp. when KIDs are used), or network based in more complex ones.

This is indeed a current limitation of the API. We would be happy to accept a PR on this.

> All those binary arguments are passed in hex encoded strings -- a conversion I'd prefer never to do on an embedded system.

[...]

PR #104, opened 2023-09-30
Allow applications to pick their C_x

chrysn:
Applications may need to pick their own C_x (as opposed to the EDHOC library picking one at random, as it was before) for three reasons:

* The library doesn't manage the set of contexts, so it can't know which C_x are already in use and which are not.
* The C_x may be used for other purposes (eg. OSCORE contexts), and might be namespaced there, which also only the application may know. (Silly example, but the application may use the OSCORE context identifiers that correspond to negative C_x for use with ACE-OSCORE profile, and only those with positive C_x for EDHOC).
* When processing of message 1 and creation of message 2 are split over time (as is necessary when implementing an EDHOC server on the coap-handler interface), it is practical to already decide on a C_R at processing time, store the responder, and dig it up when creating the response -- but that requires the C_R to be decided already at storing time.

This will likely need follow-ups for similarly delegating the choice of {ID_,}CRED_peer to the callers (resolving #99 issues from), but those are separate enough that I'd do them in a separate PR.

These are API breaking changes. They're split in two because I did C_R first, and kept them split to ease review especially on the CI side; when ACK'ed, I'd like to squash them.

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -95,12 +95,12 @@ impl<'a> EdhocResponderState<'a> {
 
     pub fn prepare_message_2(
         self: &mut EdhocResponderState<'a>,
-    ) -> Result<(BufferMessage2, u8), EDHOCError> {
+        c_r: u8,
+    ) -> Result<BufferMessage2, EDHOCError> {
         let mut cred_r: BytesMaxBuffer = [0x00; MAX_BUFFER_LEN];
         hex::decode_to_slice(self.cred_r, &mut cred_r[..self.cred_r.len() / 2])
             .expect("Decoding failed");
         let (y, g_y) = edhoc_crypto::p256_generate_key_pair();
-        let c_r = generate_connection_identifier_cbor();
 
         match r_prepare_message_2(
             self.state,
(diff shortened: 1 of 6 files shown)


malishav (approved):
This looks good to me. @geonnave do you want to take a look before merging?

lib/src/lib.rs
```diff
@@ -761,7 +770,8 @@ mod test {
         let mut initiator =
             EdhocInitiator::new(state, I, G_R, ID_CRED_I, CRED_I, ID_CRED_R, CRED_R);
 
-        let message_1 = initiator.prepare_message_1();
+        let c_i: u8 = generate_connection_identifier_cbor().into();
```
geonnave, line 773: Is `.into()` needed here (and other similar lines)? The output of `generate_connection_identifier_cbor()` is already a `u8`.

geonnave:
Thanks @chrysn, it looks good. Left a minor comment. As for merging, maybe we could merge it directly into the [remove-hacspec](https://github.com/openwsn-berkeley/edhoc-rs/tree/remove-hacspec) branch I just pushed upstream.

lib/src/lib.rs
chrysn: Whether the output of generate_connection_identifier is U8 or u8 used to depend on the back-end; those can go away when rebasing onto remove-hacspec, which is my next step.

geonnave: Oh indeed, great!

chrysn:
This is now only one commit atop #106. Rebasing was a bit of a hassle, but worked in a mixed cherry-pick and manual application approach. Looks like remove-hacspec is in a stable enough shape to rebase onto, at any rate, and things are *much* easier on that!

From my PoV, this is done once CI says Go.

geonnave:
Great! Merged into `remove-hacspec`.

PR #107, opened 2023-10-03
Add example / implementation for coap-handler

chrysn:
The coap-handler trait provides a means for servers to be implemented in a portable way across CoAP stacks. For example, it can be served from the coap crate, or from RIOT-OS. This PR adds an example of how to run an EDHOC server on there.

On the long run, this may spin into a standalone crate that pulls together EDHOC, OSCORE and CWT implementations; for the time being, it serves me as a test bed for making sure that this can be used even on very minimal CoAP implementations. I'm making this into a PR so that a) it's easier to show where requirements such as those behind #104 come from, and b) because I think it may make for a more versatile demo than the current one (at least when duly polished, at which point it'll go from draft PR into PR).

This branch is based on `remove-hacspec`, in particular because it depends on #104.

### DIFF
--- a/examples/coap/src/bin/coapserver-coaphandler.rs
+++ b/examples/coap/src/bin/coapserver-coaphandler.rs
@@ -1,0 +1,159 @@
+use edhoc_rs::*;
+use hexlit::hex;
+use std::net::UdpSocket;
+
+use embedded_nal::UdpFullStack;
+
+const ID_CRED_I: &[u8] = &hex!("a104412b");
+const ID_CRED_R: &[u8] = &hex!("a104410a");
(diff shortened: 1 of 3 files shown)


geonnave:
Feel free to target `main` now that `remove-hacspec` has been merged.

malishav:
@chrysn what is the status of this PR? It seems it passes the CI but it is still marked as a draft. Could you rebase on top of latest main?

chrysn:
Rebasing; all I'll change to replace dependency wildcards.

examples/coap/src/bin/coapserver-coaphandler.rs
```diff
@@ -0,0 +1,159 @@
...
+                // See <https://github.com/rust-embedded-community/embedded-nal/issues/47>
+                std::thread::sleep(std::time::Duration::from_millis(50));
+            }
+            e => e.expect("UDP error during send/receive"),
+        }
+    }
+}
+
+fn main() {
+    let mut stack = std_embedded_nal::Stack::default();
```
geonnave, line 157: Do you think it would make sense to rename this to `network_stack`? At first I was thinking it was the program stack 🙃.

PR #114, opened 2023-10-05
fix!: Make structs not Copy+Clone

chrysn:
Having EDHOC state structs Copy+Clone is a dangerous footgun, because it'd allow reusing a state that has already processed some message -- without the protections of the state transition checks, because they would not see a clone.

I think that this would best be extended to the low-level (proof) structs as well, but at least the way things are expressed now that's not trivial, so doing the easy and effective thing first.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -103,7 +103,7 @@ mod structs {
     }
 
     #[repr(C)]
-    #[derive(Default, Copy, Clone, Debug)]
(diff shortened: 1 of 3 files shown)


chrysn:
With the second commit on this, this is extended to the low-level structs.

This required changes to the higher-level structs' mechanisms as well, but that is really necessary: Otherwise, the user could resume after an error, which AIU is not something that should be done.

PR #127, opened 2023-11-05
refactor!: Move crypto operations into a trait.

chrysn:
This is a reroll of #103, which has become quite inconvenient to merge.

As a practical difference, the functions that used to be moved into a CryptoExt trait now just take a crypto first argument -- not much of an ergonomics setback, but easier to understand code.

PR #128, opened 2023-11-05
Typestatification

chrysn:
(Please ignore the "Move crypto operations" commit, this is just based on #127).

In the high level API (edhoc.rs is left for later, turned out to work easily that way), passing the initiator by &mut is replaced by passing it by value, and receiving a new object with different APIs in return. This ensures at build time that functions are only called when suitable -- previously this would have the low-level implementation's checks at runtime. As a side effect, the Copy+Clone derives (that could easily be used to break EDHOC's security) could be removed, obsoleting #114 bit by bit.

This picks up discussion from around https://github.com/openwsn-berkeley/edhoc-rs/issues/99#issuecomment-1748436544.

In order to make this practical, the C API is removed; it may later be reintroduced using higher level abstractions.

*Current state*: So far this only alters the initiator, not the responder. It also doesn't change  API too much. Seeing whether this breaks anything that I did not test; next steps (maybe right in this PR) is typestating the responder.


geonnave:
Thanks for this PR! I went through and it looks good to me.

geonnave:
Same comment as [here](https://github.com/openwsn-berkeley/edhoc-rs/pull/127#issuecomment-1798448319), just waiting for a hax build.

geonnave:
Same as [this comment](https://github.com/openwsn-berkeley/edhoc-rs/pull/127#issuecomment-1802816910), no issues with `hax`.

geonnave:
@malishav I am ok with merging this. Do you wish to provide any input?

PR #137, opened 2023-11-15
Refactor: re-order consts to avoid mutual recursive uses

geonnave:
This is currently a requirement of `hax`, which for now does not implement mutually recursive definitions. In practice the change re-orders some consts and flattens the `consts` and `structs` modules.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -1,257 +1,248 @@
 #![no_std]
 
+use core::marker::PhantomData;
+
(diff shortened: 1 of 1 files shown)


geonnave:
As of the current change, it doesn't complain about the consts ordering anymore.

On the other hand, this now hits one bug with hax, which I [tracked here](https://github.com/hacspec/hax/issues/360), and according to @W95Psp might be addressed today.

PR #138, opened 2023-11-15
Safe access to buffer via `.get(index)?`

geonnave:
@malishav as discussed, this is a proposal for safely accessing the received buffer. Added a partial implementation to enable a quick first-pass review.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -201,18 +201,18 @@ impl MessageBufferTrait for EdhocMessageBuffer {
 
     fn get(self, index: usize) -> Result<u8, AccessError> {
         if index >= self.len {
-            return Err(AccessError::OutOfBounds);
+            Err(AccessError::OutOfBounds)
+        } else {
+            Ok(self.content[index])
         }
-
-        Ok(self.content[index])
     }
 
     fn get_slice<'a>(&'a self, start: usize, len: usize) -> Result<&'a [u8], AccessError> {
         if start > self.len || start > len || len > self.len {
-            return Err(AccessError::OutOfBounds);
+            Err(AccessError::OutOfBounds)
+        } else {
+            Ok(&self.content[start..len])
         }
-
-        Ok(&self.content[start..len])
     }
 
     fn from_hex(hex: &str) -> Self {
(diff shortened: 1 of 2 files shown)


consts/src/lib.rs
```diff
@@ -182,6 +198,23 @@ impl MessageBufferTrait for EdhocMessageBuffer {
             len: 0,
         }
     }
+
+    fn get(self, index: usize) -> Result<u8, AccessError> {
+        if index >= self.len {
+            return Err(AccessError::OutOfBounds);
```
malishav, line 204: Why do you use `return`?

consts/src/lib.rs
```diff
@@ -154,13 +154,27 @@ impl Default for State<Start> {
...
 }
 
+pub enum AccessError {
+    OutOfBounds,
+}
+
+impl From<AccessError> for EDHOCError {
+    fn from(error: AccessError) -> Self {
+        match error {
+            AccessError::OutOfBounds => EDHOCError::ParsingError,
```
malishav, line 164: So we are basically mapping all `OutOfBounds` errors to `EDHOCError:ParsingError`. This may result in `ParsingError` for non-parsing, i.e. encoding routines when accidentally trying to access the buffer content, right?

In the case of parsing routines when unsanitized input is received, this will properly return ParsingError.

malishav (commented):
I think this is a good way forward. See inline two small comments.

consts/src/lib.rs
geonnave: Normally I prefer to halt and early-return upon errors, but I think in this case the function is small enough to use an else. Updated.

consts/src/lib.rs
geonnave: Yes, will add a TODO to think of a way to handle it in a more general case. For now it is fine since encoding routines are expected to always work (they return the value itself, and not Ok/Err).

malishav: But note that you then won't be able to use the `?` syntax and propagate the error.

consts/src/lib.rs
```diff
@@ -172,6 +187,8 @@ impl Default for EdhocMessageBuffer {
 
 pub trait MessageBufferTrait {
     fn new() -> Self;
+    fn get(self, index: usize) -> Result<u8, AccessError>;
```
chrysn, line 190: The common pattern in the standard library is that fallible access just returns an Option (eg. the slice [get function](https://doc.rust-lang.org/std/primitive.slice.html#method.get)). Why deviate here?

consts/src/lib.rs
```diff
@@ -172,6 +187,8 @@ impl Default for EdhocMessageBuffer {
...
+    fn get_slice<'a>(&'a self, start: usize, len: usize) -> Result<&'a [u8], AccessError>;
```
chrysn, line 191: The standard library uses a single [get function](https://doc.rust-lang.org/std/primitive.slice.html#method.get) that utilizes the SliceIndex trait. Could we do this here as well?

(Those two items combined might pave the way toward plugging in more widespread buffer implementations for production builds, and/or working with anything that is DerefMut<[u8]>).

consts/src/lib.rs
chrysn: That is indeed an inconvenience. The means of making `?` just turn into a `Result<..., MyError>` are [not stable yet](https://doc.rust-lang.org/std/ops/trait.FromResidual.html) (if they even work that way). If you still want to take up the suggestion, there is a [pattern](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=pub+enum+Error+%7B%0A++++A%2C%0A++++B%2C%0A%7D%0A%0Atrait+BecomesErrorB+%7B%0A++++type+Out%3B%0A++++fn+or_b%28self%29+-%3E+Result%3CSelf%3A%3AOut%2C+Error%3E%3B%0A%7D%0A%0Aimpl%3CT%3E+BecomesErrorB+for+Option%3CT%3E+%7B%0A++++type+Out+%3D+T%3B%0A++++fn+or_b%28self%29+-%3E+Result%3CT%2C+Error%3E+%7B%0A++++++++self.ok_or%28Error%3A%3AB%29%0A++++%7D%0A%7D%0A%0Afn+returning_none%28%29+-%3E+Option%3Cu32%3E+%7B%0A++++None%0A%7D%0A%0Apub+fn+parse%28%29+-%3E+Result%3Cu32%2C+Error%3E+%7B%0A++++returning_none%28%29.or_b%28%29%3F%3B%0A++++Ok%280%29%0A%7D) that allows using `foo.get(10).or_parse_error()?` in the parsing code -- although it's really just a very small enhancement over simply writing `foo.get(10).ok_or(Error::ParseError)?` every time.

geonnave: I am working on an alternative parsing mechanism, inspired by [minicbor](https://crates.io/crates/minicbor), which should solve this in the parsing functions.

geonnave:
This was superseded by #140 and #153.

PR #139, opened 2023-11-15
Crypto trait in the high level interface and EAD

chrysn:
So far, we've only used crypto traits (#127) in the low-level objects, doing the hard-to-hacspec parts. This now applies the same to the high-level objects.

It still doesn't succeed in getting the dependency on all-the-implementations out, but at least now it's just a dev-dependency (depended on by the tests), and not a full dependency any more.

Examples moved to directly using the default edhoc-crypto selection, which makes it still easy to test around; actual applications I'd expect to depend on edhoc-crypto-something directly.


chrysn:
As git dev-dependencies are [ignored by crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories), I think that with this PR we can start #98 (publication to crates.io) over in earnest: edhoc-ead*, edhoc-consts, edhoc-crypto-trait and edhoc-rs could go to crates.io; the rest will stay unpublished for the moment.

geonnave:
Went through and it looks good!

The `hax` side fails but due to a known bug, should work when [this](https://github.com/hacspec/hax/pull/364) is merged.

Will wait a bit before merge in case @malishav wants to take a look.

PR #140, opened 2023-11-16
Safe cbor parsing

geonnave:
This is an alternative approach to the safe parsing of incoming messages.

It is inspired by the [minicbor](https://gitlab.com/twittner/minicbor) library, and will enable code like this:

```rust
let input = [0x01, 0x20, 0x62, 0x68, 0x69, 0x42, 0xFE, 0xFE];
let mut decoder = CBORDecoder::new(&input);

decoder.u8()?; // 1
decoder.i8()?; // -1
decoder.str()?; // [0x68, 0x69], meaning "hi"
decoder.bytes()?; // [0xFE, 0xFE]
```

Comments:
- Now there are two layers: buffer access and CBOR decoding
  - Buffer access uses the standard `slice.get(index)?` which is safe and returns `Option`. With this we begin to be agnostic of buffer implementation.
  - Buffer access only happen inside CBORDecoder, which translates any invalid (`None`) accesses into `CBORError::DecodingError`
- The `parse_` and `decode_` functions will only interact with the `CBORDecoder`, and will not access the buffer directly.
- Errors can be chained with `?` via `impl From<CBORError> for EDHOCError`.

Finally, I also tried to use `minicbor` directly, but for the snippet above it brings an additional 20 KB of flash usage.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -190,7 +190,7 @@ pub trait MessageBufferTrait {
     fn new() -> Self;
     fn get(self, index: usize) -> Option<u8>;
     fn get_slice<'a>(&'a self, start: usize, len: usize) -> Option<&'a [u8]>;
-    fn as_slice<'a>(&'a self) -> Option<&'a [u8]>;
+    fn as_slice<'a>(&'a self) -> &'a [u8];
     fn from_hex(hex: &str) -> Self;
 }
 
@@ -213,8 +213,8 @@ impl MessageBufferTrait for EdhocMessageBuffer {
         self.content.get(start..len)
     }
 
-    fn as_slice<'a>(&'a self) -> Option<&'a [u8]> {
-        self.content.get(0..self.len)
+    fn as_slice<'a>(&'a self) -> &'a [u8] {
+        &self.content[0..self.len]
     }
 
     fn from_hex(hex: &str) -> Self {
@@ -505,11 +505,7 @@ mod edhoc_parser {
         let suites_i_len: usize;
         let mut g_x: BytesP256ElemLen = [0x00; P256_ELEM_LEN];
 
-        let mut decoder = if let Some(slice) = rcvd_message_1.as_slice() {
-            CBORDecoder::new(slice)
-        } else {
-            return Err(EDHOCError::ParsingError);
-        };
+        let mut decoder = CBORDecoder::new(rcvd_message_1.as_slice());
 
         let method = decoder.u8()?;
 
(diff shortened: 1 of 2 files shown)


chrysn:
Just to be sure that metrics are correct (and if using minicbor in earnest is on the table, which I think would be great), did you do your comparison with `lto = true`, `codegen-units = 1`, `panic = 'abort'` and `opt-level = 'z'`?

geonnave:
> Just to be sure that metrics are correct (and if using minicbor in earnest is on the table, which I think would be great), did you do your comparison with `lto = true`, `codegen-units = 1`, `panic = 'abort'` and `opt-level = 'z'`?

Yes. Cargo size gave me 105376 and 132940 for versions with/without minicbor.

The command to test: `cargo +stable size --target='thumbv7em-none-eabihf' --no-default-features --features="edhoc-crypto/cryptocell310, ead-none, rtt"`
(using stable since my mac has been complaining in the nightly build)

<details><summary>The tested consists of this change:</summary>
<p>

```patch
diff --git a/examples/edhoc-rs-no_std/Cargo.toml b/examples/edhoc-rs-no_std/Cargo.toml
index 852cfe5..9800167 100644
--- a/examples/edhoc-rs-no_std/Cargo.toml
+++ b/examples/edhoc-rs-no_std/Cargo.toml
@@ -9,6 +9,9 @@ edhoc-rs = { path = "../../lib", default-features = false }
 edhoc-crypto = { path = "../../crypto", default-features = false }
 hexlit = "0.5.3"
 
+# minicbor = "0.20.0"
+
+
 # depend on an allocator
 embedded-alloc = "0.5.0"
 
@@ -27,3 +30,9 @@ crypto-psa = [ "edhoc-crypto/psa-baremetal" ]
 crypto-cryptocell310 = [ "edhoc-crypto/cryptocell310" ]
 ead-none = [ "edhoc-rs/ead-none" ]
 ead-zeroconf = [ "edhoc-rs/ead-zeroconf" ]
+
+[profile.release]
+lto = true
+codegen-units = 1
+panic = 'abort'
+opt-level = 'z'
diff --git a/examples/edhoc-rs-no_std/src/main.rs b/examples/edhoc-rs-no_std/src/main.rs
index c8cf381..f5d5bb5 100644
--- a/examples/edhoc-rs-no_std/src/main.rs
+++ b/examples/edhoc-rs-no_std/src/main.rs
@@ -1,6 +1,6 @@
 #![no_std]
 #![no_main]
-#![feature(default_alloc_error_handler)]
+// #![feature(default_alloc_error_handler)]
 
 use cortex_m_rt::entry;
 use cortex_m_semihosting::debug::{self, EXIT_SUCCESS};
@@ -27,6 +27,8 @@ extern "C" {
     pub fn mbedtls_memory_buffer_alloc_init(buf: *mut c_char, len: usize);
 }
 
+// use minicbor::Decoder;
+
 #[entry]
 fn main() -> ! {
     #[cfg(feature = "rtt")]
@@ -51,6 +53,18 @@ fn main() -> ! {
         mbedtls_memory_buffer_alloc_init(buffer.as_mut_ptr(), buffer.len());
     }
 
+    // fn test_minicbor() {
+    //     let input = [0x01, 0x20, 0x62, 0x6F, 0x69, 0x42, 0xFE, 0xFE];
+    //     let mut decoder = Decoder::new(&input);
+
+    //     assert_eq!(1, decoder.u8().unwrap());
+    //     assert_eq!(-1, decoder.i8().unwrap());
+    //     assert_eq!("oi", decoder.str().unwrap());
+    //     assert_eq!([0xFE, 0xFE], decoder.bytes().unwrap());
+    // }
+    // test_minicbor();
+
+    
     // testing output
     println!("Hello, edhoc-rs!");
```

</p>
</details>

Issue #93, opened 2023-08-30
Proper EC point validation


malishav:
A [PR on draft-ietf-lake-traces](https://github.com/lake-wg/edhoc/pull/442) adds several invalid test vectors. Some of these contain messages with invalid G_X. In edhoc-rs, the point is not properly validated before the ECDH operation. This issue aims at adding point validation in `process_message_1` and `process_message_2` routines, which if failed, should return an error.

chrysn:
When we do this, we may want to distinguish in the type system between "this is a buffer that holds a potential public key" and "this is a buffer that holds a valid public key".

chrysn:
Proving freedom of panics (as I understand is one thing the hax converted thing can do) will be a tall order: It'll mean that the proof tool will need to remember that the only way to ever construct a public key is through the check at construction time, and that the branch that will later go from "this needs to be a valid point" to "otherwise I'll panic" will need to somehow match that knowledge.

Which, conversely, would allow incredibly defensive programming if the output of the proof could somehow be used on the Rust optimizer side to show that because there is proof of the property, the calculation doesn't have to be done again. Are there any plans to use that knowledge?

chrysn:
Seeing that this can be an attack vector to cause panics in a server, we should prioritize this; PR incoming.

PR #141, opened 2023-11-17
feat: Add rustcrypto backend

chrysn:
As it says on the tin -- this adds software based crypto backends. The added implementation is no_std itself, but using it with the out-of-thin-air constructor that the edhoc_crypto crate uses requires std (because it takes a random source from there).

There is one FIXME that I think warrants wider attention: When we get public keys passed in, who is responsible for checking that they are valid points on the curve? (So far I'm unwrapping here, which is perfectly OK if we place the responsibility in a different place, eg. in the application). This demarcation of responsibilities may well be expressed in the type system -- BytesP256ElemLen could state in its documentation that instances are always valid points (but it may make sense to use distinguished types here for "contains a public key" etc). Speaking of type system: This implementation would profit a lot if our HKDF extracts and private keys were not byte arrays but opaque types defined by the implementation. PSA will sooner or later make similar demands once people want to keep keys in a secure element. At any rate, I think those are for later.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        crypto_backend: [edhoc-crypto/hacspec, edhoc-crypto/psa]
+        crypto_backend: [edhoc-crypto/hacspec, edhoc-crypto/psa, edhoc-crypto/rustcrypto]
         ead: [ead-none, ead-zeroconf]
 
     steps:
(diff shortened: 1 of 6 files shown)


chrysn:
As a side note, this is an implementation of the crypto traits that could easily go up to crates.io, contributing to the #98 effort in that users could then start using edhoc-rs from crates.io components alone (if they choose to go with the rustcyrpto backed).

Relatedly, I don't think that this will be the implementation to end it all -- embedded users may prefer to use something hardware accelerated, while std users may prefer the assurances that can eventually come with the hacspec ones. But it's useful on embedded as a bridgehead tool, as a reference implementation, and possibly when we come to the point of enabling varied algorithms (as all relevant ones are easily available in this style).

malishav:
Wow, fabulous PR! Just quickly on your point related to point validation, this is tracked in #93. I think @geonnave was looking into it, but the basic idea is to extend the crypto trait with an additional function to verify the point, and to invoke that from message processing code.

chrysn:
Good that that's being tracked; I've changed the FIXME into a note that refers over to that other issue for justification of the possible panic.

PR #142, opened 2023-11-20
Move -crypto-trait in with -consts & related cleanups

chrysn:
When the crypto trait was introduced in #127, I tried to constrain it to the crypto/ directory. Seeing now that edhoc-consts is more and more becoming a pool of shared definitions, and keeping crates.io publication in mind, I think it's prudent to not have an extra crate, but let the crypto trait just be a module in edhoc-consts.

The refactor is marked as breaking because it removes a crate and makes a name unavailable; all affected crates are fixed by just importing the trait from somewhere else.

As bycatch now that I'm touching it, the second commit adds documentation to the trait.


chrysn:
This has turned up into a bit of a cleanup spree, all commits being documented individually. Please let me know if anything warrants better justification, I think they're all just straightforward.

geonnave:
Thanks for this PR! Looks good and fstar generation passed.

PR #144, opened 2023-11-20
Warning fixes

chrysn:
Based on #142, this removes the `#![allow(warnings)]` (which AIU is a remnant of how early hacspec required writing code that the linter would be unhappy about), and applies a lot of small refactoring fixes that would probably just have been caught at write time / when they were earlier refactored into the current state.

### DIFF
--- a/lib/src/edhoc.rs
+++ b/lib/src/edhoc.rs
@@ -45,7 +45,7 @@ pub fn edhoc_key_update(
         _th_3,
     ) = state;
 
-    let mut prk_new_buf: BytesMaxBuffer = [0x00; MAX_BUFFER_LEN];
+    let mut prk_new_buf: BytesMaxBuffer;
 
     // new PRK_out
     prk_new_buf = edhoc_kdf(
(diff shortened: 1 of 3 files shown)


chrysn:
Now ready to review (although best reviewed after #142); one warning is left (edhoc.rs line 432 about the kid never being assigned), but I suppose that's best addressed together with adding actual flexibility w/rt the peer's chosen credentials.

lib/src/edhoc.rs
```diff
@@ -45,7 +45,7 @@ pub fn edhoc_key_update(
         _th_3,
     ) = state;
 
-    let mut prk_new_buf: BytesMaxBuffer = [0x00; MAX_BUFFER_LEN];
+    let mut prk_new_buf: BytesMaxBuffer;
```
geonnave, line 48: This one we may have to revert, as `hax` does not like non-initialized variables.

chrysn: I think the preferred fix here is to not have a single buf but two bufs, not mutate them, and rely on the compiler to see that those two are not used at the same time and thus don't have to sit next to each other on the stack. Proposing the new fix (forcpushibly replacing the merge, which did both a merge and a fix at the same time)

PR #146, opened 2023-11-20
Preparations for crates.io publishing

chrysn:
This supersedes #98. In this new approach, all version related stuff is pulled together in the workspace, necessitating changes in only one file (rather than every Cargo.toml) during version bumps.

<del>This is based on #144; as GitHub still can't do staggered PRs, please start reviewing starting at "chore: pull version numbers...", and don't merge before #144.</del>


consts/Cargo.toml
```diff
@@ -1,13 +1,10 @@
 [package]
 name = "edhoc-consts"
-version = "0.1.0"
+version.workspace = true
```
geonnave, line 3: Ahh! I was wondering if there was an easy way to keep all version numbers in sync. Great!

Cargo.toml
```diff
@@ -33,6 +33,27 @@ default-members = [
 # features without this, causing forced "std" in downstream crate.
 resolver = "2"
 
+[workspace.package]
+version = "0.4.0-alpha.1"
+repository = "https://github.com/openwsn-berkeley/edhoc-rs/"
+license = "BSD"
```
chrysn, line 39: Note that this is plainly moving data around from the other crates. If you want to have #152 resolved in here right away, I'm happy to update this, but would prefer explicit confirmation in that #152 that it is to be resolved toward which-ever license (in addition to one of you merging this, which'll need to happen anyway).

malishav: Confirming BSD-3-Clause is the right license.

crypto/edhoc-crypto-rustcrypto/Cargo.toml
```diff
@@ -1,12 +1,13 @@
 [package]
 name = "edhoc-crypto-rustcrypto"
```
geonnave, line 2: @chrysn feel free to add yourself as author of this and the crypto-trait crates.

geonnave:
This looks good to me. Anything missing?

crypto/edhoc-crypto-rustcrypto/Cargo.toml
chrysn: Done (wasn't sure we're doing `authors = `, cargo doesn't set it by default any more); crypto-trait has been taken up into -consts.

chrysn:
I think this is ready. The open question of when to publish at crates.io (and on which name -- I've privately contacted the author of the edhoc crate whether he still plans to maintain that) can be answered well after this PR.

PR #147, opened 2023-11-20
Fix hacspec tests

chrysn:
Based on #144 (i.e., please merge that first and review the latest commit here), this restores the `cargo test` functionality that has not been tested recently, which https://github.com/openwsn-berkeley/edhoc-rs/issues/135 is about


geonnave:
Was merged without conflicts in #148.

PR #149, opened 2023-11-20
Refactor for hax: avoid unneeded .unwrap calls

geonnave:


consts/src/lib.rs
```diff
@@ -377,9 +377,9 @@ mod helpers {
         ))
     }
 
-    pub fn get_id_cred<'a>(cred: &'a [u8]) -> BytesIdCred {
-        let (_g, kid) = parse_cred(cred).unwrap();
-        [0xa1, 0x04, 0x41, kid]
+    pub fn get_id_cred<'a>(cred: &'a [u8]) -> Result<BytesIdCred, EDHOCError> {
+        let (_g, kid) = parse_cred(cred)?;
```
chrysn, line 381: This early return could be easily replaced with a `.map(|(_g, kid)| [0xa1, 0x04, 0x41, kid])`.

chrysn (approved):
LGTM. It replaces some unwraps with early returns, but that's probably still better from a hax PoV.

consts/src/lib.rs
geonnave: Thanks, will apply it but in #150.

PR #150, opened 2023-11-20
Refactor for hax: avoid explicit return calls

geonnave:
This is on top of #149.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -405,24 +405,23 @@ mod edhoc_parser {
         let mut ead_value = None::<EdhocMessageBuffer>;
 
         // assuming label is a single byte integer (negative or positive)
-        if let Some(b) = buffer.get(0) {
-            let label = *b;
+        if let Some((&label, tail)) = buffer.split_first() {
             let res = if CBORDecoder::is_u8(label) {
                 // CBOR unsigned integer (0..=23)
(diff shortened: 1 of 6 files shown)


consts/src/lib.rs
```diff
@@ -412,39 +412,38 @@ mod edhoc_parser {
...
-        ead_item = Some(EADItem {
-            label,
-            is_critical,
-            value: ead_value,
-        });
-        Ok(ead_item)
+            if let Ok((label, is_critical)) = res {
+                if buffer.len() > 1 {
+                    // EAD value is present
+                    let slice = &buffer[1..buffer.len()];
```
chrysn, line 422: This is replacing an explicit panic from `.get()` with an implicit one.

If you want to cut down panic options, could replace the `[1..buffer.len()]` with `[1..]` as a start, but you can also take it further and start with
```rust
if let Ok((&first, tail)) = buffer.split_first() {
    /* use first */
    /* no need for buffer.len() > 1, but go right with: */
    if Ok((&label, slice)) = tail.split_first() {
        /* use CBOR decoder on label or err out */
        /* slice here contains exactly what the annotated line panicingly  */
    }
} else {
    Err(...)
}
```

geonnave: Yes, I thought of `[1..]` on the way back home 🙃, but not the second one. Updated.

PR #153, opened 2023-11-21
Improve message buffer access

geonnave:
- [x] use `.as_slice()` instead of `buffer.content[..buffer.len]`
- [x] use `fill_with_slice(slice)` instead of `buffer[..buffer.len].copy_from_slice(slice)`

This introduces some unwrap calls which should not fail, and I hope it could be guaranteed by hax. If not then it can be reworked.

Deliberately not being exhaustive in the calls to `.content[]` within parsing routines. In this case, we also expect to have proofs from hax.

### DIFF
--- a/consts/src/lib.rs
+++ b/consts/src/lib.rs
@@ -549,7 +549,7 @@ mod edhoc_parser {
         let id_cred_r: IdCred;
         let mut mac_2: BytesMac2 = [0x00; MAC_LENGTH_2];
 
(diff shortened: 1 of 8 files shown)


chrysn:
By the way, the MessageBuffer is at this point pretty much a re-implementation of [heapless::Vec](https://docs.rs/heapless/latest/heapless/struct.Vec.html). If hacspec grows well, it may be possible to just swap that in at some point. (That'd work best when all direct access to content has been removed, which AIU we currently do for efficiency reasons -- thus, it may take some time until we can do that, for may require placement return).

PR #155, opened 2023-11-23
chore(CI): Run hax always, using pre-built docker image

chrysn:
Let's see if building is as fast as I'd expect now that there is a hax image on dockerhub.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -74,7 +74,7 @@ jobs:
       run: |
         docker run --rm -v ${{ github.workspace }}:/edhoc-rs docker.io/geonnave/hax bash -e -c "
         cd edhoc-rs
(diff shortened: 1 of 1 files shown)


geonnave:
Wow, container pull + hax gen in 2 minutes!

chrysn:
I think this is ready now. Please have an extra look at whether the removal of ead-none is right here. Possibly, we may want to add another line that does the check on the consts module while it is behaving weirdly as you described in chat -- but that's probably nothing that'll stop this PR.

chrysn:
I think this is ready now. Please have an extra look at whether the removal of ead-none is right here. Possibly, we may want to add another line that does the check on the consts module while it is behaving weirdly as you described in chat -- but that's probably nothing that'll stop this PR.

geonnave:
The removal of `ead-none` was right. I pushed a change to reflect the following:
- we are not typechecking the crypto backends
- we want to typecheck edhoc-consts
- call hax with edhoc-consts after edhoc-rs, just in case

PR #157, opened 2023-11-24
Changes after hax meeting

geonnave:
- avoid using match bindings
- fix buffer access bug


geonnave:
I now realize that `hax` sometimes does not yield errors, even though the generated `.fstar` files may contain messages such as `something is not implemented yet`. So, maybe, in addition to calling `cargo-hax` and making sure it doesn't error, we should also ensure that the return of `find -name *.fst -type f | xargs grep "something is not implemented yet"` is empty.

PR #160, opened 2023-11-30
Refactor: make state struct use named fields

geonnave:


geonnave:
In the future we may want to have different structs for each phase, but for now this at least allows us to avoid many de-structuring calls as well as the nasty `state.6` thing.

PR #165, opened 2023-12-06
Rename to `lakers`

chrysn:
Picking "lakers" as discussed in today's call with @malishav and @geonnave. This PR only renames the crates that'd actually go to crates.io, and uses 

* "edhoc-rs" => "lakers"
* "edhoc-consts" => "lakers-shared" (because it's not really constants any more; that's a pending rename anyway)
* "edhoc-crypto-rustcrypto" => "lakers-crypto-rustcrypto" (which so far is the only backend that doesn't need patching and can go up as a publicly usable reference)
* "edhoc-ead" => "lakers-ead-dispatch" (edhoc-crypto would be renamed to edhoc-crypto-dispatch too, because that's what they do)
* "edhoc-ead-none" => "lakers-ead-none"
* "edhoc-ead-zeroconf" => "lakers-ead-zeroconf"

All the renaming is for now purely local to the Cargo.toml files; inside Rust no identifiers are replaced. We'll probably want to do that at some point in time, but that's best done on a flag day when nobody has any pending PRs or (worse) local changes, as that's a big search-and-replace (and may make sense together with directory renaming etc).

I'll probably need one or two iterations on CI to fix mentions of names I haven't found in local testing.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -38,8 +38,8 @@ jobs:
     - name: Run unit tests with default features
       run: RUST_BACKTRACE=1 cargo test
 
-    - name: Run unit tests with feature matrix # note that we only add `--package edhoc-ead-zeroconf` when testing with that config
-      run: RUST_BACKTRACE=1 cargo test -p edhoc-rs -p edhoc-crypto -p edhoc-consts ${{ matrix.ead == 'ead-zeroconf' && '-p edhoc-ead-zeroconf' || '' }}  --no-default-features --features="${{ matrix.crypto_backend }}, ${{ matrix.ead }}" --no-fail-fast -- --test-threads 1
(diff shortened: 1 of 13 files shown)


geonnave:
This looks good to me. `lakers-shared` is indeed more aligned with its current functionality, and `ead-dispatch` too. I think I am missing the reason for not renaming `edhoc-crypto` to `lakers-crypto-dispatch`.

chrysn:
> I think I am missing the reason for not renaming `edhoc-crypto` to
> `lakers-crypto-dispatch`.

edhoc-crypto, and the other non-renamed crates, are just not part of the
initial upload. When renaming everything, I wouldn't be sure where to
stop (like, examples too?), so I've picked a minimal set necessary for
the task at hand.

geonnave:
Sounds good!

PR #166, opened 2023-12-06
Prepare 0.4.0 alpha.1

chrysn:
Experimenting with crates.io uploads I'm still finding subtleties of the criteria that are applied to crates before they can be published.

This includes the changes from https://github.com/openwsn-berkeley/edhoc-rs/pull/165 as it builds on it.

<del>Testing whether any of them affect CI here, will switch to reviewable once all passes both here and crates.io's tests.</del>

I've published the lakers crates v0.4.0-alpha.1 from this branch. @geonnave and @malishav, I'd like to set you as crate owners (precisely: add you, then it's up to you if you'll still have me around there) on crates.io, but that'll only work if you log in to https://crates.io/ through GitHub once to make yourselves known there.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -38,8 +38,8 @@ jobs:
     - name: Run unit tests with default features
       run: RUST_BACKTRACE=1 cargo test
 
-    - name: Run unit tests with feature matrix # note that we only add `--package edhoc-ead-zeroconf` when testing with that config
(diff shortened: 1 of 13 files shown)


geonnave:
Heyy it's cool to see `lakers` at docs.rs and https://crates.io/crates/lakers 😄, thanks for pushing this!

I've just logged into crates.io with my GitHub account as requested.

geonnave:
This looks good to me.

PR #170, opened 2023-12-08
Refactor EADs to not use global state

geonnave:
This is a step in the direction of the separate message processing, initially discussed in #99, and further discussed in the meeting this week.

Edit: this includes:
- no more global mutable state
- no more worrying about having the same API on every EAD handler (in fact I don't think it is possible without `dyn` or some crazy generic stuff)
- better code organization, including typestated device (U) and authenticator (V)


geonnave:
The new `lakers-ead-authz` crate is ready by itself, but now it depends on `lakers` to be updated to support multi-step message processing. Thus for now I'll keep it as _draft_.

geonnave:
@chrysn I tried to use the typestate pattern here, could you please take a look? (see the 2 commits with "typestatify" in the name")

chrysn:
The typestatification looks about right (as far as I can tell that from mainly looking at types and signatures). It looks like there are fewer types than there were enum states before, but apparently not all were used. Is ZeroTouchAuthenticatorState left around on purpose?

geonnave:
Thank you for the review. 

> apparently not all (types) were used

Yes, some were unused and the `Error` state is now simply a variant of `Result`.

I can't find the symbol `ZeroTouchAuthenticatorState`, but maybe you are referring to the whole `lakers-ead-zeroconf` crate, which I did not remove yet because it breaks tests. I will remove once the split is merged, which uses the changes in this PR for the EAD handing.

Issue #94, opened 2023-09-12
Stateful initialization of crypto backends


geonnave:
In the `cryptocell310` backend, I changed the initialization code so that it only runs once after boot.
This reduced handshake time by ~80 milliseconds.
Here is the commit for that: https://github.com/geonnave/edhoc-rs/commit/d58b2ae2d9aa029925af7718fe7ad87516931875

The goal with this issue is to bring that change to other crypto backends.

chrysn:
As you pointed out, this could benefit from https://github.com/openwsn-berkeley/edhoc-rs/pull/103

I don't think it'd even need actual fields in the Crypto -- it could still be ZST, but it'd be only created through the initializer function:

```rust
#[derive(Copy, Clone)]
pub struct Crypto {
    _private: ()
}

/// Initialize the cryptography engine by doing some pre-computations.
///
/// unsafe: This writes to a static mut, and thus may only be called while no
/// [Crypto] struct exists. In particular, it is safe to call once, as this is the only
/// constructor of a Crypto.
unsafe fn edhoc_rs_crypto_init() -> Crypto {
    unsafe { /* one-time initialization stuff */ };
    Crypto { _private: () }
}
```

(If we want to do without unsafe, that's just as fine -- we can have a `struct Crypto(CRYS_RND_State_t)` with a safe constructor, and `impl CryptoTrait for &Crypto`, but with the unsafe C SaSi_LibInit etc functions around, it's probably easier this way.)

PR #174, opened 2023-12-12
Split message processing

geonnave:
On top of PR #170. Aims to provide the split message processing discussed in issue #99.

Main tasks:
- [x] split message processing
- [x] some cleanup of the values stored in the outer state
- [x] ~~update to only pass cred_x later on (instead of when calling `initiator/responder::new`~~ EDIT: tracked in #179
- [x] update the examples

In addition, as a result of the split there is now `IdCred` and `IdCredOwned`. Thus,
- [x] ~~this should be made homogeneous.~~ EDIT: tracked in #178.

Finally, with respect to the EAD:
- [x] integrate the refactored `lake-authz` EAD implementation
- [x] ~~implement `add_eadX` functions~~ EDIT: tracked in #177


shared/src/lib.rs
```diff
@@ -142,6 +142,34 @@ pub struct State<Phase: EDHOCState> {
...
+pub struct PartialMessage1 {
+    pub c_i: u8,                    // connection identifier chosen by the initiator
+    pub x_or_y: BytesP256ElemLen,   // ephemeral private key of myself
+    pub gx_or_gy: BytesP256ElemLen, // g_y or g_x, ephemeral public key of myself
+    pub suites_i: BytesSuites,
+    pub suites_i_len: usize,
+}
+
+#[derive(Debug)]
+pub struct WaitMessage2New {
```
chrysn, line 155: AIU WaitMessage2New replaces the State<WaitMessage2>. That's a welcome replacement -- can the WaitMessage2 be removed now already?

chrysn (commented):
A few comments inline.

From a final reviewing PoV, I'd hope to see the refactoring in shared and lib separated from the addition of an application, but that can be a split in the PR when it's going out of the draft stage (as keeping them split during development may incur excessive rebasing).

shared/src/lib.rs
```diff
@@ -91,7 +91,7 @@ pub type EADMessageBuffer = EdhocMessageBuffer; // TODO: make it of size MAX_EAD
 // This is sealed
 pub trait EDHOCState: core::fmt::Debug {}
 // For both initiator and responder
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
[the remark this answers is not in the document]
geonnave: It's just so that we can do `EdhocInitiator::new(Default::default(), ...)` in the tests.

shared/src/lib.rs
geonnave: Yep, it is WIP and thus a bit of a mess for now; in the end I will remove all the old stuff.

geonnave:
The split part works. It still requires:
- [ ] some cleanup of the values stored in the outer state
- [ ] update to only pass cred_x later on (instead of when calling `initiator/responder::new`
- [ ] update the examples

In addition, as a result of the split there is now `IdCred` and `IdCredOwned`. Thus,
- [ ] this should be made homogeneous.

Finally, I should
- [ ] integrate the refactored `lake-authz` EAD implementation

geonnave:
This is ready for an initial review round. Note that the split begins at [5dd6c72](https://github.com/openwsn-berkeley/lakers/pull/174/commits/5dd6c722556d043170a9b5110de7bf059b334d85) (previous commits are from PR #170).

chrysn:
This looks to me like it's really independent of #170 (apart from commits such as 49cff1a2e13f501fb8e38ff262a9770cb0c75cee that applies small fixes).

I'm not sure how you do / want reviews done here, but to me this would be way easier to review if the splitting changes could happen in a branch of their own. Depends a bit on the level of review you want here -- if it's more a rough look you're after, that works here as well. If you prefer commit-by-commit review, that might be more productive on a dedicated branch.

shared/src/lib.rs
```diff
@@ -127,37 +100,76 @@ pub enum EDHOCError {
...
 }
 
-#[repr(C)]
 #[derive(Debug)]
-pub struct State<Phase: EDHOCState> {
-    pub current_state: PhantomData<Phase>,
-    pub x_or_y: BytesP256ElemLen,   // ephemeral private key of myself
-    pub c_i: u8,                    // connection identifier chosen by the initiator
-    pub gy_or_gx: BytesP256ElemLen, // g_y or g_x, ephemeral public key of the peer
+pub struct Start;
```
chrysn, line 104: How long does generating a key take? (I have no clue, can be just-take-the-few-bytes-of-entropy, can be do-some-complex-GF2-computations). If it's not trivial, it may be a good idea to store the x_or_y already in the Start, as that allows a system to prepare a key while not in use.

(Could also spin that out to an own low-prio issue)

lib/src/edhoc.rs
```diff
@@ -288,53 +341,66 @@ pub fn r_process_message_3(
...
+        gx_or_gy: g_x,
+        suites_i,
+        suites_i_len,
+    })
+}
 
+pub fn i_prepare_message_1b(
+    state: PartialMessage1,
+    crypto: &mut impl CryptoTrait,
+    ead_1: &Option<EADItem>, // FIXME: make it a list of EADItem
```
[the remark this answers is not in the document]
chrysn: If we went with that style (`add_ead1(...)`), we might also take it further to do
```rust
let (state, msg1) = start.prepare_message1(x, g_x, c_i)
    .add_ead1(id, Some(bytes))?
    .add_ead1(more_id, None)?
    .finish(&mut crypto);
```
but as that's a larger refactoring, maybe keeping the way to add a single option here is a practical way forward, and we can do less-splitting-more-refactoring changes still later.

chrysn (commented):
Still mid-review, leaving some comments.

For context, the glass through which I'm looking at the code right now is `git diff --ignore-all-space 5dd6c72^..geonnave/split-message-processing -- shared lib`.

shared/src/lib.rs
geonnave: I only have data for preparing message 1, which as a whole takes about 20 ms, given these conditions: crypto initialisation is stateful (see #94); using the cc310 backend. So it is fast, although software backends might take longer.

geonnave:
Thanks for the review so far. The type of review I am looking for is about the current state of the API and an overall rough look.

lib/src/edhoc.rs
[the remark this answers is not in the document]
geonnave: Agreed that it can be addressed separately. Tracked in #177.

chrysn:
I think the rough API is good from a high-level PoV. There are lots of details we can enhance, but we can do that progressively once this is in.

One thing probably makes sense to do now because it's a pain to do later: look through the newly introduce type and function names. (Now this can be one commit with `grep -i s/.../.../ **/*.rs`, later that's bound to cause merge conflicts). Particular candidates are the split functions that are now `a` and `b` variants. The process_message_n / prepare_message_n+1 split already works well, so maybe process_message_3a / process_message_3b can become parse_message_3 / verify_message_3, and prepare_message_3a / preparere_mesage_3b can become prepare_message_3 / finalize_message_3 (same for prepare_message_1a/b)

geonnave:
Did changes suggested by @malishav (un-split prepare_message_1 and 3) and the renamings proposed by @chrysn. 
Another critical item pointed by @malishav (private key accessed by application) is tracked in #182 and will be fixed ASAP.

PR #175, opened 2023-12-12
doc: Add README for shared, and link READMEs in the crates

chrysn:
This should fix the big looming "appears to have no README.md file" texts on crates.io. I don't think we can really test this other than wait what it looks like after the next release, but the absense of "unused manifest key" warnings in a cargo check is a good sign.

The main crate just uses the top-level README; all others use a shared one that points to the main crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ resolver = "2"
 version = "0.4.1"
 repository = "https://github.com/openwsn-berkeley/edhoc-rs/"
 license = "BSD-3-Clause"
+readme = "shared/README.md"
 
 [workspace.dependencies]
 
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -37,6 +37,7 @@ resolver = "2"
 version = "0.4.1"
 repository = "https://github.com/openwsn-berkeley/edhoc-rs/"
 license = "BSD-3-Clause"
+readme = "shared/README.md"
```
geonnave, line 40: why it points to the README of the shared crate?

geonnave:
Thank you for this PR, it looks good. Left one minor comment.

chrysn:
> why it points to the README of the shared crate?

Because it fast-forwards the user to the relevant main crate with the
overview documentation. Several crates with all the same text (and the
user won't know unless they diff, maybe the detail is somewhere down)
are confusing to users. The shared README is ... well, shared among all
other crates :-) and is just that: a pointer that says it's part of the
big thing.

geonnave:
I see, then it can just be overridden when crate-specific READMEs are added. Cool!

PR #176, opened 2023-12-12
doc: Add a few high-level introductory descriptions

chrysn:
Building on #175, this makes the docs.rs pages a bit prettier, mainly adding top-level docs and some text about the most frequently used structs.

If you agree with the text in there this one should be easy and self-contained. I'll eventually want to add more, but leading there will be a round of revisiting whether a great many things really need to be public. (No harm in documenting private items, but public items are priorized higher).


geonnave:
Went through and looks good!

PR #183, opened 2023-12-26
Fix: application no longer accesses private ephemeral key

### LINKED ISSUE #182: Avoid accessing private ephemeral key from application code
The private key was accessed by the application after the split-message and new EAD architecture PRs.

geonnave:
This fixes #182.


malishav (commented):
I think this is the way to go. Just a note that the name `compute_secret` in `src/lib.rs` can be misleading since there are multiple secrets that the library can expose. I would suggest using a more explicit name, e.g. `compute_ephemeral_secret`, which should stress the fact that the ECDH secret is computed from the ephemeral private key.

Issue #179, opened 2023-12-21
Pass CRED_X later on


geonnave:
Currently the identity selected for the handshake at the `Start` state. This could be made more flexible, in particular for the initiator:
- at the initiator side, it could choose which identity to use in the handshake after learning CRED_R, i.e., after `process_message_2a`
- at the responder, I am not sure it makes a difference, as it discloses its identity before learning CRED_I. Nevertheless, the responder's identity is really only needed at `prepare_message_2`

geonnave:
Solved in #186.

PR #186, opened 2024-01-10
Set initiator identity after learning R's

geonnave:
Solves #179.

I think it only makes sense to apply this to the initiator side, as the responder has to select an identity without knowing who is the initiator anyway (cc @malishav @chrysn).


malishav (approved):
I think that this change makes sense. I agree with your assessment that it doesn't make much sense on doing the same on R side.

Issue #187, opened 2024-01-10
Implement "credential transfer" option


geonnave:
Ability for Initiator/Responder to explicitly set whether the credential should be sent by value or by reference.

geonnave:
Solved in #188

PR #188, opened 2024-01-10
Explicitly set credential transfer option

geonnave:
Addressing #187. 

With this:
1. no more need to pass information (kid or full credential) already present in cred_x (which already is in the state)
2. the choice of how to transfer the credential becomes more clear


geonnave:
@malishav since this changes the API can you give a quick look? (see commit [af6cc25](https://github.com/openwsn-berkeley/lakers/pull/188/commits/af6cc25bab578bae0d4514fe70eaf19f4eed2ae4))

PR #189, opened 2024-01-10
Use owned credential type

geonnave:


geonnave:
My initial goal was to get rid of `IdCredOwned`, and replace it with `CredentialRPK`, but then fields such as `CredentialRPK.value` would have to be `Optional`, spreading around runtime checks that would otherwise not be needed.

So I've decided to just keep `IdCredOwned` as is.

Thus, the main change in this PR is the addition of the `CredentialRPK` type which owns its contents and provide static access to fields such as `public_key` and `id_cred`, thus removing one potential source of panics from the protocol core.

Issue #177, opened 2023-12-21
Ability to prepare and process more than one EAD field


geonnave:
Currently one either 0 or 1 EAD fields can be added to a message. This is not complete according to the EDHOC RFC, which specifies 0 or N EAD fields.

From the API perspective, here is one possible way to implement it (see [this review comment](https://github.com/openwsn-berkeley/lakers/pull/174#discussion_r1425394633)):

```rust
let (state, msg1) = start.prepare_message1(x, g_x, c_i)
    .add_ead1(id, Some(bytes))?
    .add_ead1(more_id, None)?
    .finish(&mut crypto);
```

chrysn:
If we started this now I'd probably rename a few things and make that

```rust
let (state, msg1) = start.prepare_message1(x, g_x, c_i)
    .with_ead(id, Some(bytes))?
    .with_ead(more_id, None)?
    .finish(&mut crypto);
```

Given that the message is built in a buffer, I also hope that we can do without allocating per-EAD buffers, and copy all the items into the message as it is being built. With self->self functions, this should be fine with hax.

PR #192, opened 2024-01-11
Add EAD field to MAC computation

### LINKED ISSUE #191: Include EAD in MAC computation, when present
I just realised that `EAD_1` and `EAD_2` are missing from `MAC_2` and `MAC_3`, respectively. The EAD items should be passed to `compute_mac_x` [as an optional item](https://www.ietf.org/archive/id/draft-ietf-lake-edhoc-22.html#name-responder-composition-of-me) , e.g., `context_2 = << ID_CRED_R, TH_2, CRED_R, ? EAD_2 >>`

geonnave:
Fixes #191

### DIFF
--- a/lib/src/edhoc.rs
+++ b/lib/src/edhoc.rs
@@ -115,6 +115,7 @@ pub fn r_prepare_message_2(
         &cred_r.get_id_cred(),
         cred_r.value.as_slice(),
         &th_2,
+        ead_2,
     );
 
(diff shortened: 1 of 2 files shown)


geonnave:
This adds another field to the core state (`ead_X`). It could be solved by instead passing an offset to the EAD item in the plaintext, but I think that it is better to solve it when tacking #177.

Issue #180, opened 2023-12-21
Re-enable C API


geonnave:
From version [0.2.0](https://github.com/openwsn-berkeley/lakers/releases/tag/REL-0.2.0) to [0.3.0](https://github.com/openwsn-berkeley/lakers/releases/tag/v0.3.0) the library had support for being compiled as a static library + automatic generation of C headers, enabling seamless integration with C projects. From [0.4.0](https://github.com/openwsn-berkeley/lakers/releases/tag/v0.4.0) onwards, however, the API changed, including typestatification and (currently ongoing) split message preparation/processing. 

While things were being moved around, it made sense to temporarily remove the C API. This issue aims at re-enabling it, so that `lakers` can also be made available for the embedded C world.

geonnave:
Solved in #197

PR #197, opened 2024-01-17
Re-add C API, a.k.a. `lakers-c`

geonnave:
Addressing #180.

Since now the crypto backends are dissociated from the main crate, I had to add a new, separate crate that will import `lakers` and `lakers_crypto`. One good implication of this approach is that this new crate can be hardcoded to `crate-type = ["staticlib"]` (in the previous version, I had to manually change from "rlib" to "staticlib" everytime I wanted to build the static version). On the other hand, I still need to figure whether `cbindgen` will actually work across the `lakers_ffi, lakers, and lakers_shared` crates and generate all the needed headers.


geonnave:
Generating structs pulling from other crates in the workspace is fine, one challenge now is how to work around the generic `crypto` item in the structs found in `lib.rs`.

geonnave:
This is ready for review / merge. For an overview on the API and the integration with `authz`, see `lakers-c-native/main.c` and the changes in `coap/coapserver.rs`.

examples/lakers-c-native/main.c
```diff
@@ -0,0 +1,199 @@
...
+    print_hex(payload, edhoc_msg_len+1);
+    coap_add_data(pdu, edhoc_msg_len + 1, payload);
+    // coap_show_pdu(COAP_LOG_WARN, pdu);
+    if (coap_send(session, pdu) == COAP_INVALID_MID)
+    {
+        coap_log_err("cannot send CoAP pdu\n");
+        return -1;
+    }
+    while (has_coap_response == 0)
+        coap_io_process(ctx, COAP_IO_WAIT);
```
malishav, line 73: please add curly brackets even though there is a single instruction within the loop, strong opinion :)

PR #219, opened 2024-02-13
Add CI action to build python wheels for many platforms

geonnave:


geonnave:
It works! It is now only enabled for when a release tag is published. See an example execution at https://github.com/openwsn-berkeley/lakers/actions/runs/7902018726

PR #231, opened 2024-03-04
Add cpp_compat for C headers

### LINKED ISSUE #230: Make default C headers C++ compatible
When lakers-c headers are generated by the `cbindgen` tool they currently are strictly C compatible, meaning cannot be linked to with C++. But there is a cbindgen option [`cpp_compat`](https://github.com/mozilla/cbindgen/blob/f1d5801d3b299fa2e87d176f03b605532f931cb6/docs.md?plain=1#L472) which enables C++ compatibility in a way which can still be C compatible.

The default lakers configuration in [cbindgen.toml](https://github.com/openwsn-berkeley/lakers/blob/main/lakers-c/cbindgen.toml) should include this option in the top-level group:
```
cpp_compat = true
```

BrianSipos:
This affects the headers for the lakers_shared and lakers_c libraries.

Closes #230

### DIFF
--- a/lakers-c/cbindgen.toml
+++ b/lakers-c/cbindgen.toml
@@ -10,6 +10,7 @@ header = """
  */"""
 include_guard = "LAKERS_C_H"
 includes = ["lakers_shared.h", "lakers_ead_authz.h"]
+cpp_compat = true
 
 [parse.expand]
 all_features = true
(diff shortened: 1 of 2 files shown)


geonnave:
Looks good, thank you!

PR #235, opened 2024-03-06
examples/coap: Update to new traits

chrysn:
This vastly enhances error handling.

This is still a PR because I think that EDHOCError should become renderable, so that EDHOC errors can fly out through `?`.

### DIFF
--- a/examples/coap/src/bin/coapserver-coaphandler.rs
+++ b/examples/coap/src/bin/coapserver-coaphandler.rs
@@ -20,6 +20,28 @@ struct EdhocHandler {
     connections: Vec<(u8, EdhocResponderWaitM3<Crypto>)>,
 }
 
+/// Render a MessageBufferError into the common Error type.
+///
+/// It is yet to be determined whether anything more informative should be returned (likely it
+/// should; maybe Request Entity Too Large or some error code about unusable credential.
+///
+/// Places using this function may be simplified if From/Into is specified (possibly after
+/// enlarging the Error type)
+fn too_small(_e: MessageBufferError) -> Error {
+    Error::bad_request()
+}
+
+/// Render an EDHOCError into the common Error type.
+///
+/// It is yet to be decided based on the EDHOC specification which EDHOCError values would be
+/// reported with precise data, and which should rather produce a generic response.
+///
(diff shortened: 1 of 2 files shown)


chrysn:
Enhancing around the handling of EDHOCError is probably a more gradual thing, so I'm marking this as ready for review as it is (yes there are FIXMEs in the code, but those are still better than the panics that were in their place in the last iteration).

examples/coap/src/bin/coapserver-coaphandler.rs
```diff
@@ -2,6 +2,8 @@ use hexlit::hex;
 use lakers::*;
 use lakers_crypto::Crypto;
 
+use coap_message::{Code, MinimalWritableMessage, MutableWritableMessage, ReadableMessage};
+use coap_message_utils::{Error, OptionsExt as _};
```
geonnave, line 6: `Error` is very generic, maybe alias to something like `CoapError`?

geonnave:
Looks good, left two comments.

examples/coap/src/bin/coapserver-coaphandler.rs
```diff
@@ -53,49 +77,73 @@ enum EdhocResponse {
 
 impl coap_handler::Handler for EdhocHandler {
     type RequestData = EdhocResponse;
-    fn extract_request_data(
+
+    type ExtractRequestError = Error;
```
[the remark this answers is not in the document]
chrysn: This is not an alias, this is an associated type of a trait.

examples/coap/src/bin/coapserver-coaphandler.rs
chrysn: It's the only Error type this code touches; I'd prefer not to name types around.

On the long run, I expect that this module grows its own error type (that is a union of coap_message_utils::Error and the two wrapped errors), and implements the CoAP error rendering. Then, that will be called Error, and there will be a From<coap_message_utils::Error> to it (with coap_message_utils::Error explicitly named rather than imported because it will be rare by then). But so far, every kind of error encountered in this program is eventually converted into this type.

geonnave: Sounds good.

PR #236, opened 2024-03-06
shared: Remove numeric value, repr(C) and UnknownError from EDHOCError

chrysn:
The numbers associated to EDHOCError are confusing because they do not match the EDHOC ERR_INFO numeric values. Let's try whether we really need them.

(Frankly, at this stage this is looking for whether CI takes it).

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -107,15 +107,15 @@ pub type EADMessageBuffer = EdhocMessageBuffer; // TODO: make it of size MAX_EAD
 #[repr(C)]
 #[derive(PartialEq, Debug)]
 pub enum EDHOCError {
-    UnknownPeer = 1,
-    MacVerificationFailed = 2,
-    UnsupportedMethod = 3,
-    UnsupportedCipherSuite = 4,
-    ParsingError = 5,
-    EadLabelTooLongError = 6,
-    EadTooLongError = 7,
-    EADError = 8,
-    UnknownError = 9,
+    UnknownPeer,
+    MacVerificationFailed,
+    UnsupportedMethod,
+    UnsupportedCipherSuite,
+    ParsingError,
+    EadLabelTooLongError,
+    EadTooLongError,
+    EADError,
+    UnknownError,
 }
 
 #[derive(Debug)]
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -120,7 +120,15 @@ pub enum EDHOCError {
     ParsingError,
     EadLabelTooLongError,
     EadTooLongError,
-    EADError,
+    /// An EAD was received that was either not known (and critical), or not understood, or
+    /// otherwise erroneous.
+    EADUnprocessable,
+    /// The credential or EADs could be processed (possibly by a third party), but the decision
+    /// based on that was to not to continue the EDHOC session.
+    ///
+    /// See also
+    /// <https://datatracker.ietf.org/doc/html/draft-ietf-lake-authz#name-edhoc-error-access-denied>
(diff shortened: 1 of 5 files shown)


chrysn:
Hm, first pass worked. From chats it seems repr(C) is not essential there anyway, and the UnknownError is not used (and rather bad practice in error types compared to non_exhaustive); adding two more simple (semver breaking, but so is removal of the numeric values) changes.

geonnave:
I said it on chat but I was wrong: let's keep `repr(C)`. Right now I don't use it but it is good to make `EDHOCError` available for implementations that need to check the error.

The numeric value can stay omitted, though.

chrysn:
Shouldn't the `#[repr(C)]` be added if and when the type is exposed to C, given the C side can't make use of the data without copying constants? And I think that the preferred abstraction would be to have an `edhoc_error_get_err_id()` function depending on how things go with aligning with the EDHOC spec. (I'll update to repr(C) if you persist, just wanted to mention the alternative before.)

chrysn:
Added an accessor as per chat. This is not yet mapped to C, but can be (but so far, the repr(C) on the error type was not used either).

shared/src/lib.rs
```diff
@@ -104,18 +106,72 @@ pub type BytesMac = [u8; MAC_LENGTH];
...
+    /// If an EDHOCError is returned through EDHOC, it will use this in its EDHOC error message.
+    ///
+    /// Note that this on its own is insufficient to create an error message: Additional ERR_INFO
+    /// is needed, which may or may not be available with the EDHOCError alone.
+    ///
+    /// TODO: Evolve the EDHOCError type such that all information needed is available.
+    pub fn err_code(&self) -> ErrCode {
+        use EDHOCError::*;
+        match self {
+            UnknownPeer => ErrCode::UNKNOWN_CREDENTIAL,
```
geonnave, line 139: If I parsed [EDHOC Section 6.4](https://www.ietf.org/archive/id/draft-ietf-lake-edhoc-23.html#section-6.4) correctly, UNKNOWN_CREDENTIAL would only be used when the credential is sent by reference.
In the implementation, UnknownPeer is used in both cases: by value and by reference.
So maybe add a note mentioning this.

shared/src/lib.rs
```diff
@@ -104,18 +106,72 @@ pub type BytesMac = [u8; MAC_LENGTH];
...
+            MacVerificationFailed => ErrCode::UNSPECIFIED,
+            UnsupportedMethod => ErrCode::UNSPECIFIED,
+            UnsupportedCipherSuite => ErrCode::WRONG_SELECTED_CIPHER_SUITE,
+            ParsingError => ErrCode::UNSPECIFIED,
+            EadLabelTooLongError => ErrCode::UNSPECIFIED,
+            EadTooLongError => ErrCode::UNSPECIFIED,
+            EADError => ErrCode::ACCESS_DENIED,
```
geonnave, line 146: This is not going to always be ACCESS_DENIED. 
Ideally would be like this:
- EADError would map to UNSPECIFIED
- a new EADAccessDeniedError to ACCESS_DENIED (returned from handle_voucher_request)

For now I think it should map to UNSPECIFIED.

chrysn: Addressed by splitting the error into EADUnprocessable and AccessDenied (which may come from an authz process, but just as well from credentials processing, just that we don't implement ACLs for credentials at the moment)

shared/src/lib.rs
geonnave: Sounds good, thanks.

PR #237, opened 2024-03-06
Remove lakers-ead crate

chrysn:
I don't think the lakers-ead crate is necessary any more: Since EADs are now driven by the application, they can be additive (one build test can contain them all). The test matrix was reduced as part of this because the tested applications did not actually pull code from the crate that toggled its content (couldn't either, because crates don't see feature flags of dependencies, they can only set them).

I probably missed some spots in the code during the cleanup, waiting for CI to show me where.

(Pleas only look at the last commit; the rest is actually #235, which was my starting point)


geonnave:
I think this change makes sense.

chrysn:
Apparently some of the tests don't actually test on the current commit. As a remedy, rebasing onto current main (and squashing in the process).

chrysn:
CI passed, & squashed. I think it's ready.

geonnave:
Looks good, thank you.

PR #239, opened 2024-03-07
coapserver-coaphandler: Restore feature parity with coapserver

chrysn:
authz support was added to the coapserver but not to coapserver-coaphandler; this fixes the regression.

This is based on #238 because the C client's bugs were hidden by coapserver's lax handling (see https://github.com/martindisch/coap-lite/issues/36), and would otherwise make it unusable for even basic testing.


geonnave:
Looks good, thanks for this PR.

PR #241, opened 2024-03-07
examples: Add traits based CoAP example

chrysn:
Unlike the server, this is using embedded-nal-async, because that's what a CoAP implementation exists for that can do client as well. (I may update the server side to use that as well instead of the blocking version, but they don't conflict).

Having a server and a client that work with embedded CoAP stacks is an important step towards having a pair that can exchange requests over 6LoWPAN on embedded hardware (apart from the executor setup and possibly the random source, both should work on RIOT just fine).

PR marked as WIP as I'd like to have authz in there, and am struggling with it. (The only working client side example I have is in C, and that appears to do the static-static derivation manually from reaching deep down into the state, but that can't be it, given how smoothly the server side works).

### DIFF
--- a/examples/coap/Cargo.toml
+++ b/examples/coap/Cargo.toml
@@ -19,3 +19,12 @@ coap-handler = "0.2"
 coap-handler-implementations = "0.5"
 coap-numbers = "0.2.3"
 coap-message-utils = "0.3.1"
+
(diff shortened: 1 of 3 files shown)


geonnave:
This looks good (although I didn't run it), and the authz code seems to be there. Anything else missing from your point of view?

PR #243, opened 2024-03-13
Fixing hkdf_expand in crypto/psa

### LINKED ISSUE #242: crypto/psa: logic error on hkdf_extract
This was the actual source of the bug identified in #234.

TL;DR: the `hkdf_expand` function in the `edhoc-crypto/psa` backend generates a wrong `keystream`, it only happens when the key length is long enough, and can only be detected when different backends are used by initiator and responder.

To understand the bug, consider [the call](https://github.com/openwsn-berkeley/lakers/blob/95677a26374d09dd1b26e14784a03d7dcd05e5d5/lib/src/edhoc.rs#L898) to `edhoc_kdf` in `encrypt_decrypt_ciphertext_2`. The bug will happen if `ciphertext.len > 32` (details below). There are two reasons why this bug hasn't been caught before:
1. most test vectors and examples use a small enough ciphertext_2, due to sending credentials by reference. however, when sending a credential by value then the ciphertext length is larger, e.g. 96 bytes in my particular test.
2. the bug only shows up when using different backends in the initiator and responder, i.e., if `psa` is used on both sides, both are equally incorrect so the bug cancels out.

## Samples

For example here is the generated `keystream_2` with:

1. the `rustcrypto` backend: 
```
 [17, 190, 198, 253, 100, 121, 147, 41, 196, 149, 210, 171, 230, 53, 165, 125,
 163, 185, 0, 8, 218, 196, 92, 35, 45, 151, 72, 143, 58, 131, 211, 19, 193, 101,
 80, 111, 113, 16, 184, 125, 112, 89, 145, 211, 128, 183, 159, 190, 215, 151,
 101, 205, 192, 199, 96, 144, 136, 106, 171, 224, 187, 207, 138, 169, 173, 172,
 155, 2, 95, 26, 60, 95, 12, 98, 12, 213, 223, 104, 144, 166, 3, 155, 161, 176,
 251, 2, 94, 171, 119, 197, 142, 157, 220, 201, 120, 134, 0, <many omitted zeros
 follow>]
```

2. the `psa` backend: 
``` [115, 167, 251, 228, 4, 176, 40, 121, 91, 166, 230,
247, 81, 60, 147, 171, 143, 141, 171, 214, 235, 223, 143, 112, 101, 157, 145,
68, 1, 245, 122, 247, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 238, 129, 15, 159, 48, 124, 106, 19, 204,
15, 20, 217, 45, 17, 40, 77, 40, 195, 87, 200, 221, 159, 46, 195, 141, 42, 175,
128, 129, 75, 132, 248, 0, <many omitted zeros follow>] 
```

Notice the section with many zeroes in the `psa` version -- that's the bug. And that is why, in #234, only the first 32 bytes of the ciphertext were being correctly decrypted.

## How to fix

The fix is to adjust [this for loop](https://github.com/openwsn-berkeley/lakers/blob/95677a26374d09dd1b26e14784a03d7dcd05e5d5/crypto/lakers-crypto-psa/src/lib.rs#L63) in `hkdf_expand` so that the generated keystream will be properly computed and saved into the output array.

Unit tests with vectors from [RFC5869](https://datatracker.ietf.org/doc/html/rfc5869) must be included as well in the fix.

geonnave:
Fixes #242

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -29,6 +29,7 @@ rand_core = { version = "0.6.4", optional = true, default-features = false }
 
 [dev-dependencies]
 hexlit = "0.5.3"
+rstest = "0.11.0"
 
 [features]
 default = [  ]
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -27,6 +27,9 @@ lakers-crypto-cryptocell310 = { workspace = true, optional = true }
 lakers-crypto-rustcrypto = { workspace = true, optional = true }
 rand_core = { version = "0.6.4", optional = true, default-features = false }
 
+[dev-dependencies]
+hexlit = "0.5.3"
+
 [features]
 default = [  ]
 # hacspec = [ "lakers-crypto-hacspec" ]
--- a/crypto/lakers-crypto-psa/src/lib.rs
+++ b/crypto/lakers-crypto-psa/src/lib.rs
@@ -60,13 +60,13 @@ impl CryptoTrait for Crypto {
         let mut t_i = self.hmac_sha256(&message[..info_len + 1], prk);
         output[..SHA256_DIGEST_LEN].copy_from_slice(&t_i);
 
-        for i in 2..n {
+        for i in 2..=n {
             message[..SHA256_DIGEST_LEN].copy_from_slice(&t_i);
             message[SHA256_DIGEST_LEN..SHA256_DIGEST_LEN + info_len]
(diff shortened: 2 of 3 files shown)


malishav (approved):
Thanks for providing the PR and finding the bug! Looks good.

PR #246, opened 2024-03-15
testing: re-enable hacspec backend

geonnave:
It was previously disabled since it was failing to build on CI (external reason).

This is a test to see if that still happens.

If CI no longer fails, the backend can be re-enabled.


geonnave:
The issue persists, therefore closing this.

Issue #256, opened 2024-04-19
Python: parse_message_2 produces id_cred_r as array


chrysn:
The 2nd return value of parse_message_2, id_cred_r, currently has values like `[10]` or `[162, 2, 96, 8, 161, 1, 165, 1, 2, 2, ...]`. What it represents is the binary value of ID_CRED_R, a byte string, and it should be represented as such in Python.

(Haven't had a look at the Python implementation yet, might manage a PR.)

chrysn:
Something similar was fixed in 0e2b07bae3f39ae5048cf17269664f35a81658ee already. I think the general rule should be that binary data in return values should not be Vec<u8> but PyBytes. (In function arguments it can be convenient to have them as Vec<[u8]> from an API PoV; from a performance PoV it may just as well be a &[u8] if that works). PR available, let's see if CI likes it.

PR #257, opened 2024-04-19
python: Don't return Vec<u8>

chrysn:
The ID_CRED_x returned in the initiator and responder APIs are unintuitive for Python users when arrays (esp. if they see examples where it's a `[10]` style single value -- to a Python user that looks more like a parsed CBOR, more so if that is actually what parsing a CBOR sequence would give). Returning it as Python bytes makes it easier to understand that this is the opaque data indicating either an identifier or a credential by value.

Closes: https://github.com/openwsn-berkeley/lakers/issues/256

### DIFF
--- a/lakers-python/src/initiator.rs
+++ b/lakers-python/src/initiator.rs
@@ -57,19 +57,20 @@ impl PyEdhocInitiator {
         }
     }
 
-    pub fn parse_message_2(
+    pub fn parse_message_2<'a>(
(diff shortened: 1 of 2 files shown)


geonnave:
Looks good, thank you. (this was an overlook on my side, initially did all with Vec<u8> but then realized bytes was better, and obviously forgot to update some functions 🙃)

Issue #258, opened 2024-05-03
C_x only support u8


chrysn:
We currently only support C_x as u8. Given that the peer may choose a longer one, this is an interoperability hazard.

As far as I have found, there is no limit to their length in EDHOC, so the implementation will need to pick a limit, which can be reasonably 1, but right now this is not configurable.

chrysn:
As a first step, I'll start wrapping the u8 in a newtype that'll allow changing this later.

PR #260, opened 2024-05-03
C_x: Represent using own ConnectionIdentifier newtype

chrysn:
This makes the APIs clearer and paves the way for adding support for longer C_x.

Contributes-To: https://github.com/openwsn-berkeley/lakers/issues/258

Marked as a draft while waiting for CI checks (I only did `cargo test` which usually misses some)

### DIFF
--- a/examples/coap/src/bin/coapclient.rs
+++ b/examples/coap/src/bin/coapclient.rs
@@ -51,7 +51,7 @@ fn client_handshake() -> Result<(), EDHOCError> {
     let valid_cred_r = credential_check_or_fetch(Some(cred_r), id_cred_r).unwrap();
     let initiator = initiator.verify_message_2(&I, cred_i, valid_cred_r)?;
 
-    let mut msg_3 = Vec::from([c_r]);
+    let mut msg_3 = Vec::from(c_r.as_cbor());
     let (mut initiator, message_3, prk_out) =
         initiator.prepare_message_3(CredentialTransfer::ByReference, &None)?;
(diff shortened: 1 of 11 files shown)


geonnave:
Thanks, so far looks good.
Evolving it into N-sized identifiers would imply some changes in the encoder/decoder (for which you seem to have a plan as well).
Would not oppose calling the new type `ConnId`.

chrysn:
Thanks. Big rename coming up with the squash. Yes, I have a plan, and I'm adding a `.from_slice()` constructor that'll allow at least the Python API to be nice from the start (so that you pass in bytes, get out bytes, and it just complains when the bytes contain more than 1 byte, of a value unsupported so far -- then, this can progress transparently). For the C API, nothing.

geonnave:
Great!

> For the C API, nothing.

That concerns me a bit, tracking at #262.

chrysn:
> For the C API, nothing.

More precisely, things *do* work, but for the same reasons I advocated against having the C API the last time it was removed I don't want so sink too much time there, so I am fixing the examples, but not the API; consequently, unlike the Python API, once we do support longer identifiers, the C API can't use them.

chrysn:
All green, squashed and ready for final review.

geonnave:
Looks good, thanks.

I can take on updating the C API later -- there's an important C-based project in the lab which need lakers working on it.
Still this way it helps a lot, in that we both share parts of the work to advance the library.

PR #261, opened 2024-05-03
Expose C_I to responders

chrysn:
Depends-On: #260 (therefore, just check the top commit)

Closes: #258

On the Python side, this follows the convention established in #260 to represent ConnId as bytes instead of an integer (so we won't need an API change there once ConnId can get larger)


chrysn:
All green, squashed and ready for review!

geonnave:
Good! Thanks.

PR #264, opened 2024-05-03
shared/ConnId: Fix logic error in constructor

chrysn:
This fixes a copy-paste error stemming from using code from from_int_raw without inverting the conditions, and adds documentation including doctests on this and related methods.

---

This fixes a bug introduced in #260 that was not covered by tests (because the tests do not try the path where the initiator deliberately chooses a C_I)

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -132,6 +132,23 @@ impl ConnId {
     }
 
     /// The CBOR encoding of the identifier.
+    ///
+    /// For the 48 compact connection identifiers -24..=23, this is identical to the slice
+    /// representation:
+    ///
+    /// ```
+    /// # use lakers_shared::ConnId;
+    /// let c_i = ConnId::from_slice(&[0x04]).unwrap();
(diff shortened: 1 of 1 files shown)


shared/src/lib.rs
```diff
@@ -132,6 +132,23 @@ impl ConnId {
     }
 
     /// The CBOR encoding of the identifier.
+    ///
+    /// For short the 48 compact connection identifiers, this is identical to the slice
```
geonnave, line 136: > For short the 48 compact connection identifiers

I don't understand this part.

geonnave: Maybe just replace it by `For compact connection identifiers (integer value in -24..=23)`. Referring to the range is easier to track to the RFC if needed.

geonnave:
Looks good, left one comment, thanks.

shared/src/lib.rs
chrysn: Good point, fix forcepushed.

PR #267, opened 2024-05-14
lib: Allow ID_CRED_I to be a value

chrysn:
Currently, EDHOC panics when attempting to use use a credential by value in CRED_I.

This fixes this, mainly by copying code from the CRED_R case where that was already supported.


chrysn:
Note that this is not quite trivial to test because using the test vector credential by value (it is relatively long given we use p256 and not x25519, and on top of that has a serial number in it) exceeds the 256 default length by some bytes.

chrysn:
I think this is showing a few cases where instead of doing CBOR encoding there is the tacit assumption that a byte value is just CBOR_MAJOR_BYTE_STRING ^ value :-/

chrysn:
I've added a crude CBOR dissection for the message 3 lengths, and adjusted the Python tests so that different variations are run. I didn't quickly find where else we had that kind of CBOR processing -- refactoring would be welcome, but getting this out with the next release might be more urgent than doing this cleanly.

lib/src/edhoc.rs
```diff
@@ -638,17 +641,18 @@ fn edhoc_kdf(
...
     ead_3: &Option<EADItem>,
 ) -> Result<BufferPlaintext3, EDHOCError> {
     let mut plaintext_3: BufferPlaintext3 = BufferPlaintext3::new();
 
     // plaintext: P = ( ? PAD, ID_CRED_I / bstr / int, Signature_or_MAC_3, ? EAD_3 )
-    plaintext_3.content[0] = id_cred_i[id_cred_i.len() - 1]; // hack: take the last byte of ID_CRED_I as KID
-    plaintext_3.content[1] = CBOR_MAJOR_BYTE_STRING | MAC_LENGTH_3 as u8;
-    plaintext_3.content[2..2 + mac_3.len()].copy_from_slice(&mac_3[..]);
-    plaintext_3.len = 2 + mac_3.len();
+    id_cred_i.write(&mut plaintext_3)?;
```
geonnave, line 651: This gives the impression that `id_cred_i` is being written on. Would something like the code below work?

```
plaintext_3.extend_from_slice(id_cred_i.as_cbor())
```

Or else just `s/write/write_to_message/`

lib/src/edhoc.rs
```diff
@@ -744,11 +759,22 @@ fn decrypt_message_3(
...
+    let prefix_length;
+    // FIXME: Reuse CBOR decoder
+    if (0..=23).contains(&(message_3.content[0] ^ CBOR_MAJOR_BYTE_STRING)) {
+        bytestring_length = (message_3.content[0] ^ CBOR_MAJOR_BYTE_STRING).into();
+        prefix_length = 1;
+    } else {
+        // FIXME: Assumes we don't exceed 256 bytes which is the current buffer size
+        bytestring_length = message_3.content[1].into();
+        prefix_length = 2;
+    }
```
geonnave, line 772: Same as above.

lib/src/edhoc.rs
```diff
@@ -723,16 +727,27 @@ fn encrypt_message_3(
...
+    if bytestring_length < 24 {
+        output.content[0] = CBOR_MAJOR_BYTE_STRING | (bytestring_length) as u8;
+        prefix_length = 1;
+    } else {
+        // FIXME: Assumes we don't exceed 256 bytes which is the current buffer size
+        output.content[0] = CBOR_MAJOR_BYTE_STRING | 24;
+        output.content[1] = bytestring_length as _;
+        prefix_length = 2;
+    };
+    output.len = prefix_length + bytestring_length;
```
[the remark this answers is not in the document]
chrysn: I think that the version with assigning to the variable once is more readable -- it makes it clear immediately that the 1 and 2 is the prefix length to the reader, and the compiler not only makes the same out of it, it even (by virtue of this being uninitialized and immutable) ensures it is assigned exactly once.

lib/src/edhoc.rs
chrysn: See https://github.com/openwsn-berkeley/lakers/pull/267#discussion_r1601247211 -- can change it if you insist, but it was a conscious choice. Here, the alternative is `let (bytestring_length, prefix_length) = match { ... (message_3.content[1].into(), 2) ...}` that is even less clear to the reader -- that was actually what prompted me to use the assign-once variables in both cases.

lib/src/edhoc.rs
```diff
@@ -853,20 +879,10 @@ fn encode_plaintext_2(
...
-        }
-        IdCred::FullCredential(cred) => {
-            plaintext_2.content[c_r.len()] = CBOR_BYTE_STRING;
-            plaintext_2.content[c_r.len() + 1] = cred.len() as u8;
-            plaintext_2.content[c_r.len() + 2..][..cred.len()].copy_from_slice(cred);
-            c_r.len() + 2 + cred.len()
-        }
-    };
+    plaintext_2.extend_from_slice(c_r).unwrap();
+    id_cred_r.write(&mut plaintext_2)?;
```
geonnave, line 884: Same as above.

lib/src/edhoc.rs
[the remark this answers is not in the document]
chrysn: As for the assert, I'm adding one now (although the slicing would panic later anyway), but on the long run I think we'll want way more fallible access in the style of `write()` and less manual manipulation of indices and lengths -- hax should manage with those more readable patterns by now.

lib/src/edhoc.rs
```diff
@@ -853,20 +879,10 @@ fn encode_plaintext_2(
...
-            c_r.len() + 1
-        }
-        IdCred::FullCredential(cred) => {
-            plaintext_2.content[c_r.len()] = CBOR_BYTE_STRING;
-            plaintext_2.content[c_r.len() + 1] = cred.len() as u8;
-            plaintext_2.content[c_r.len() + 2..][..cred.len()].copy_from_slice(cred);
-            c_r.len() + 2 + cred.len()
-        }
-    };
+    plaintext_2.extend_from_slice(c_r).unwrap();
```
geonnave, line 883: I know that the `copy_from_slice` could already lead to panicking behavior, but we could improve it and avoid the explicit `unwrap` here by:

```diff
diff --git a/lib/src/edhoc.rs b/lib/src/edhoc.rs
index d29e01b..b793258 100644
--- a/lib/src/edhoc.rs
+++ b/lib/src/edhoc.rs
@@ -880,7 +880,7 @@ fn encode_plaintext_2(
     let mut plaintext_2: BufferPlaintext2 = BufferPlaintext2::new();
     let c_r = c_r.as_slice();
 
-    plaintext_2.extend_from_slice(c_r).unwrap();
+    plaintext_2.extend_from_slice(c_r).or(Err(EDHOCError::EncodingError))?;
     id_cred_r.write(&mut plaintext_2)?;
     let offset_cred = plaintext_2.len;
 
diff --git a/shared/src/lib.rs b/shared/src/lib.rs
index 9e29870..5bee193 100644
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -193,6 +193,7 @@ pub enum EDHOCError {
     UnsupportedMethod,
     UnsupportedCipherSuite,
     ParsingError,
+    EncodingError,
     CredentialTooLongError,
     EadLabelTooLongError,
     EadTooLongError,
@@ -229,6 +230,7 @@ impl EDHOCError {
             UnsupportedMethod => ErrCode::UNSPECIFIED,
             UnsupportedCipherSuite => ErrCode::WRONG_SELECTED_CIPHER_SUITE,
             ParsingError => ErrCode::UNSPECIFIED,
+            EncodingError => ErrCode::UNSPECIFIED,
             CredentialTooLongError => ErrCode::UNSPECIFIED,
             EadLabelTooLongError => ErrCode::UNSPECIFIED,
             EadTooLongError => ErrCode::UNSPECIFIED,
```

lib/src/edhoc.rs
chrysn: Renamed in 0f2fef8994262c988bc54690a32f5e1a84f96200

lib/src/edhoc.rs
chrysn: LGTM, applied.

lib/src/edhoc.rs
geonnave: Sounds good.

geonnave:
Yep good on my side. Merging as CI passes.

Issue #269, opened 2024-05-15
Python: Construct CredentialRPK from parts


chrysn:
Right now, any CredentialRPK usable through the Python API is constructed through `::new()` (and thus through `::parse()`, which recognizes some shapes of CCSes and nothing more).

Even if our parse became much better, recognized any valid CCS and X509 on top of it, credentials that are CWTs could not be processed without a lot of additional context.

So let's allow Python users to construct a CredentiaRPK by having a CWT (which they would pass in by value) and the corresponding public key extracted from it (which they would do by decrypting it, eg. using pycose, or because it is the public part of their own private key, in which case they may not even be able to decrypt the CWT, but would know what is supposed to be in there because they got it for their own key).

Chances are we'll have a Python wrapper for CredentialRPK with a constructor `CredentialRPK(credential, publickey=None)`, and maybe we can even stay API compatible by allowing the credential to be passed in as bytes where it is accepted as bytes already, in which case it will be converted just as it was.

(Not sure if I'll get around to implementing it)

chrysn:
Should have known I couldn't leave it be :-/

chrysn:
Attempting to *use* this quickly runs into trouble in other places where the credential being parseable is required (eg. r_parse_message_3), but I hope that this still useful enough on its own that we can start brainstorming how to proceed from this being merged.

PR #270, opened 2024-05-15
 Python: Construct CredentialRPK from parts

chrysn:
Closes: https://github.com/openwsn-berkeley/lakers/issues/269


geonnave:
Thanks, just went through it. Looks like it allow constructing from either just value or public key and kid? (or maybe value and public key and kid?)

I see that the API doesn't change, but maybe there could be a new test showing the new supported behavior?

geonnave:
Looks good, thanks.

Issue #272, opened 2024-05-16
MACed ID_CRED_x is always {4: "k"}


chrysn:
Currently, MAC calculations always generates a {4: "k"} where k is the 1-long string encoded in the numeric kid:

```rust
    let mac_2 = compute_mac_2(
        crypto,
        &prk_3e2m,
        c_r,
        &cred_r.get_id_cred(),
        cred_r.value.as_slice(),
        &th_2,
        ead_2,
    );

    pub fn get_id_cred(&self) -> BytesIdCred {
        [0xa1, 0x04, 0x41, self.kid] // cbor map = {4: kid}
    }
```

However, as I understand the RFC, the value should be
* what we currently do iff the serialized ID_CRED_x was a number
* something similar to what we do currently on longer KID (which we don't support)
* the literal ID_CRED_x otherwise

This is only becoming apparent through the latest PRs #267, #270 and #271 -- without those, pass-by-kid was the only option anyway.

I think we'll have to overhaul a bit how we transport ID_CRED_x, CRED_x and how we link them to each other.

I anticipate that any plug test we do with a different implementation will fail as soon as we try anything but ID_CRED_x-is-KID with another implementation.

chrysn:
I propose we differentiate CredentialRpk into two types (this is clearly an API change):

* Credential similar to CredentialRpk, but without the kid, and it always has a value and a key
* IdCredential, which is either an enum (u8 | bytes (not initially) | CBOR item in a buffer), or just a CBOR item in a buffer (I think the preferred serialization there is the compact one, but whatever we choose, building the corresponding other form is either along the lines of "prepend a10441 to it if it doesn't start with a0/3", or of "if it starts with a10441, strip those 3 bytes").

By being different types, the using application is forced to do a conversion. Currently we do that conversion implicitly by one of two ways:
1. If it is by-value and we can parse it, we parse it
2. If it is by-reference and matches what we configured earlier, we take that

I don't like that we do 2 anyway (it is asking the application to set up something that it expects way earlier than it has the information), but if we want to stick with 2, we can still allow it in the API by having a combined parse-or-recognize method; I'd personally rather get rid of it and just have a fallible parse method (that recognizes an IdCred to be by value and then turns it into a Cred) and a match or lookup method, and run those independently of the context.

geonnave:
[...]

Issue #273, opened 2024-05-16
Are we doing credential-by-value wrong?


chrysn:
The places where ID_CRED_x is used in the plaintexts is described in RFC9528 as `ID_CRED_R / bstr / -24..23`.

My understanding is that fundamentally, ID_CRED_x is always a 1-element map, and that the bstr / -24..23 are just the shorthands of [Section 3.5.3.2](https://datatracker.ietf.org/doc/html/rfc9528#compact-kid) that expand to `{4: bstr}` (where -24..23 are also translated to bstr, and regular bstr are used as-is). The RFC also registers dedicated keys for that map for kccs and kcwt (we currently only use CCSs).

What we send and parse when doing credential-by-value is that we take the CRED_x value (which is a byte string), encode it as a CBOR byte string, and include it in the plain text. This was recently refactored into IdCred::write_to_message in #267 (while not altering Lakers' behavior).

I believe that this is an error, and that instead of using 482E2E2E2E2E2E2E2E ('........') we should be using A10E2E2E2E2E2E2E2E2E ({/kccs/ 14: '........'}). The RFC9529 traces do not contain an instance of credential-by-value that I could find. Has sending a credential by value ever been plugtested?

PR #274, opened 2024-05-16
Introduce Credential and IdCredential as successors to CredentialRpk

chrysn:
Closes: #272, #273

This is a first sketch of how I'd go about #272; the design includes some choices influenced by my assessment of #273. The two new types should eventually replace CredentialRpk completely (and also remove the need for BytesIdCred).

If I'm wrong about #273, Credential may still have a `.value_type`, but it wouldn't be needed for `.by_value()` any more, only for `.by_thumbprint()` (which we don't have yet).

@geonnave, is this a good direction?

### DIFF
--- a/shared/src/cred.rs
+++ b/shared/src/cred.rs
@@ -70,6 +70,165 @@ impl CredentialRPK {
     }
 }
 
+/// A credential along with its corresponding public key
+///
+/// It may contain information about the key ID, which enables its use by reference, and of its
+/// value type, which enables its use by value.
+///
+/// Future extensions may also add other optional identifiers (eg. an identifying URI) that might
+/// enable additional kinds of referncing the credential; some of those may also not be stored but
+/// merely generated (eg. thumbprints, although in that case the type of credential needs to be
+/// stored, so that it can be decided whether the thumbprint is an x5t or a c5t).
+#[derive(Clone, Copy, Debug)]
+pub struct Credential {
+    /// Bytes of the credential. Lakers and EDHOC make no requirement on their format.
+    value: EdhocMessageBuffer,
+    /// Public authentication key (G_I or G_R) expressed in the credential
+    // For CCSs, this could be a reference into value, but Rust does not allow self-referential structs anyway.
+    public_key: BytesP256ElemLen,
+    /// Key ID.
+    ///
+    /// If set, it enables turning the credential into an IdCred that is a key ID (the most compact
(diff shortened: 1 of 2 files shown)


chrysn:
This needed updates after understanding the need for #281.

shared/src/cred.rs
```diff
@@ -70,6 +70,158 @@ impl CredentialRPK {
...
+    /// form of ID_CRED_x).
+    ///
+    /// Future versions may widen this to allow longer KIDs.
+    kid: Option<u8>,
+    /// COSE Header Parameter Label indicating the type of credential this is.
+    ///
+    /// If set, it enables using the credential by value. Setting this adds the requirement that
+    /// the value is well-formed according to the type. In particular, when set to KCCS or any
+    /// other Map-valued types, a value that is not CBOR can result in hard to debug parsing errors
+    /// on the other end, even before it reaches the credential validation at the peer.
```
geonnave, line 119: s/at the peer/at the peer's application/

geonnave:
Yes I think this is in the right direction.
Not really sure how `ByValueStyle` will be used, but I think it will become clearer as it progresses.
Thanks for the contributions so far.

shared/src/cred.rs
```diff
@@ -70,6 +70,158 @@ impl CredentialRPK {
...
+/// enable additional kinds of referncing the credential; some of those may also not be stored but
+/// merely generated (eg. thumbprints, although in that case the type of credential needs to be
+/// stored, so that it can be decided whether the thumbprint is an x5t or a c5t).
+#[derive(Clone, Copy, Debug)]
+pub struct Credential {
+    /// Bytes of the credential. Lakers and EDHOC make no requirement on their format.
+    value: EdhocMessageBuffer,
+    /// Public authentication key (G_I or G_R) expressed in the credential
+    // For CCSs, this could be a reference into value, but Rust does not allow self-referential structs anyway.
+    public_key: BytesP256ElemLen,
```
geonnave, line 106: can you help me understand why the value and public_key are not optional, while the kid is optional?

for context, I will have to add support for an experimental PSK credential (for the emerging edhoc-psk draft), and I am trying to see how things are going to fit here. for example, I was thinking that the PSK can be stored separately (like the private keys), and the credential type might hold only the kid that identifies the PSK.

geonnave:
This PR served very well as inspiration to the credential-related updates in #284! Since that one is now merged, I am closing this.

PR #279, opened 2024-05-22
Fix credential-by-value

### LINKED ISSUE #273: Are we doing credential-by-value wrong?

geonnave:
Fixes #273.

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -62,10 +62,13 @@ pub const CBOR_MAJOR_BYTE_STRING: u8 = 0x40u8;
 pub const CBOR_MAJOR_BYTE_STRING_MAX: u8 = 0x57u8;
 pub const CBOR_MAJOR_ARRAY: u8 = 0x80u8;
 pub const CBOR_MAJOR_ARRAY_MAX: u8 = 0x97u8;
+pub const CBOR_MAJOR_MAP: u8 = 0xA0;
 pub const MAX_INFO_LEN: usize = 2 + SHA256_DIGEST_LEN + // 32-byte digest as bstr
 				            1 + MAX_KDF_LABEL_LEN +     // label <24 bytes as tstr
 						    1 + MAX_KDF_CONTEXT_LEN +   // context <24 bytes as bstr
 						    1; // length as u8
+
+pub const KCSS_LABEL: u8 = 14;
 
 pub const ENC_STRUCTURE_LEN: usize = 8 + 5 + SHA256_DIGEST_LEN; // 8 for ENCRYPT0
 
@@ -506,8 +509,14 @@ impl<'a> IdCred<'a> {
             IdCred::FullCredential(cred) => {
                 let len =
                     u8::try_from(cred.len()).map_err(|_| EDHOCError::CredentialTooLongError)?;
+                let kccs_map_len = 1;
                 message
-                    .extend_from_slice(&[CBOR_BYTE_STRING, len])
+                    .extend_from_slice(&[
+                        CBOR_MAJOR_MAP + kccs_map_len,
+                        KCSS_LABEL,
+                        CBOR_BYTE_STRING,
+                        len,
+                    ])
                     .map_err(|_| EDHOCError::CredentialTooLongError)?;
                 message.extend_from_slice(cred)
             }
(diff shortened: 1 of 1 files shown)


examples/coap/src/bin/coapclient.rs
```diff
@@ -53,7 +53,7 @@ fn client_handshake() -> Result<(), EDHOCError> {
 
     let mut msg_3 = Vec::from(c_r.as_cbor());
     let (mut initiator, message_3, prk_out) =
-        initiator.prepare_message_3(CredentialTransfer::ByReference, &None)?;
+        initiator.prepare_message_3(CredentialTransfer::ByValue, &None)?;
```
chrysn, line 56: Do we want to change this here or is this just an artifact of debugging?

shared/src/lib.rs
```diff
@@ -49,7 +49,7 @@ pub const ENCODED_VOUCHER_LEN: usize = 1 + MAC_LENGTH; // 1 byte for the length
 // maximum supported length of connection identifier for R
 pub const MAX_KDF_CONTEXT_LEN: usize = 150;
 pub const MAX_KDF_LABEL_LEN: usize = 15; // for "KEYSTREAM_2"
-pub const MAX_BUFFER_LEN: usize = 256;
+pub const MAX_BUFFER_LEN: usize = 256 + 32;
```
chrysn, line 52: Is this intentional or just an artifact of debugging with credentials that happen to be a bit longer than needed? (A CRED_x that does not carry the full dash separated MAC address may be short enough)

shared/src/lib.rs
```diff
@@ -937,6 +950,19 @@ mod cbor_decoder {
...
 
+        /// Begin decoding a map.
+        pub fn map(&mut self) -> Result<usize, CBORError> {
+            let b = self.read()?;
+            if CBOR_MAJOR_MAP != Self::type_of(b) {
+                Err(CBORError::DecodingError)
+            } else {
+                match Self::info_of(b) {
+                    31 => Err(CBORError::DecodingError), // no support for unknown size arrays
+                    n => Ok(self.as_usize(n)?),
```
chrysn, line 961:
```suggestion
                    n if n < 24 => Ok(self.as_usize(n)?),
                    31 => Err(CBORError::DecodingError), // no support for long or indeterminate size
```

chrysn (commented):
LGTM, didn't test yet. Comments inline.

chrysn:
It appears that something is wrong with the commits. Could it be that your changes were committed --amend to the latest merge commit?

chrysn:
I've taken the liberty to force-push a version that has an actual commit in there.

chrysn:
I have now tested this and it works on self-interop. Let's see if we can even squeeze one more interop test into this hackathon.

shared/src/lib.rs
geonnave: Accepted.

examples/coap/src/bin/coapclient.rs
geonnave: Accepted.

shared/src/lib.rs
geonnave: Solved

chrysn:
We're not there yet: from interop, we did `{/kccs/14: <<{...}>>}` while Marco (and the IANA registry) say it is `{/kccs/14: {...}}`

geonnave:
Final part of the fix provided in PR #281.

PR #281, opened 2024-05-23
CCS: Process as map rather than CBOR encoded map

chrysn:
Following up on #273, this hopefully brings the credentials by value in line with the specification.

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -1031,11 +1031,12 @@ mod cbor_decoder {
                         _ => unreachable!("Value was masked to 5 bits"),
                     };
                     match major {
-                        0 | 1 | 7 => (), // Argument consumed, remaining items were already decremented
+                        0..=1 => (), // Argument consumed, remaining items were already decremented
+                        7 => (), // Same, but in separate line due to Hax FStar backend limitations
                         6 => {
                             remaining_items += 1;
                         }
-                        2 | 3 => {
+                        2..=3 => {
                             self.read_slice(argument.into())?;
                         }
                         4 => {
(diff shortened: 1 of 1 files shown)


chrysn:
I've tested this between Lakers devices. No plugtest happend on this solution yet, but I think it's enough of an improvement to merge it once reviewed -- I don't know when I next get around to doing more interop tests with others.

examples/coap/src/bin/coapclient.rs
```diff
@@ -53,7 +53,7 @@ fn client_handshake() -> Result<(), EDHOCError> {
 
     let mut msg_3 = Vec::from(c_r.as_cbor());
     let (mut initiator, message_3, prk_out) =
-        initiator.prepare_message_3(CredentialTransfer::ByReference, &None)?;
+        initiator.prepare_message_3(CredentialTransfer::ByValue, &None)?;
```
geonnave, line 56: Maybe this and the `+ 32` below were pulled in by mistake from a previous commit?

chrysn: Kind of -- I hadn't pulled before working from your branch, now following the rebase.

shared/src/lib.rs
```diff
@@ -967,12 +988,77 @@ mod cbor_decoder {
...
+            // of this (especially not keep looping needlessly) and doesn't do anything worse than
+            // keep a limited loop counter.
+            for _ in self.buf.iter() {
+                if remaining_items > 0 {
+                    remaining_items -= 1;
+                    let head = self.read()?;
+                    let major = head >> 5;
+                    let minor = head & 0x1f;
+                    let argument = match minor {
+                        0..=23 => minor,
```
geonnave, line 1016: Shouldn't we use all numeric values in hex (to stay consistent with the rest of the parser)?

chrysn: The rest of the parser usually operates on full bytes (major | additional), which we have the constants for. This is the odd one out because it is operating on them separately, and when those values are referenced in RFC8949, it is always by numeric value ("major type 6", "additional information / less than 24".

I can convert if you prefer.

geonnave: I see, sounds good then.

geonnave:
Looks good to me! Leave a :+1: here when you think it is ready to merge.

Issue #280, opened 2024-05-23
Debug ergonomics


chrysn:
Currently, our errors are coarse-grained (which is fine), and are returned as results rather than panicking (which is also fine).

A consequence of this is that it is very hard to trace down where an error occurs, because you can't just set break points to the constructor of (say) an EDHOCError::ParsingError, so without a time-travelling debugger and without std printf, it is very hard to find where parsing actually failed. (For example, to find https://github.com/openwsn-berkeley/lakers/pull/279#issuecomment-2125201223, I added a ParsingErrorS(&'static str) variant and changed all potential construction sites to include a label).

Thing is, I don't know a good way to do that so that we can have the labels around. The labels I'd like to have are not intended as content for the error messages, merely as debug output (winding up in the impl Debug of the errors), and on production builds, those should be elided at compile time. None of the options seem to work right:
1. If we add a `('static str)` content to all the error variants, we'll have to add a label to *all* constructors (much work but OK), but we can't compile time disable it with a `#[cfg()]` unless we also cfg-gate the arguments in the constructors (tedious)
2. If we change EDHOCError to be a `struct { EDHOCErrorVariant, Option<'static str> }`, we'll have to construct them through the internal variant as `EDHOCErrorVariant::ParsingError.with_label("foo")`, need to add a `.without_label()` or `.into()` whenever we directly return errors (on `?` paths we can skip it and use the automatic `.into()`), and the match branches grow more verbose.
3. We could simplify 2 a bit if the EDHOCError had associated constants `EDHOCError::ParsingError = EDHOCError(EDHOCErrorVariant::ParsingError, None)`, but that requires duplicating the variants across both types, requires a clippy override to allow an associated constant to use enum variant naming conventions, and will be super confusing in match because you can list all the variants and then the compiler complains about unhandled variants.

So, the issue is a bit annoying (esp. during plug testing), but I like none of the solutions I can come up with. Do you have better ones?

geonnave:
Yeah, normally I only do "hot debug", i.e. `rust-gdb --tui target/debug/coapclient`, but that needs to be interactive.

I think the simple and good enough solution would be a logging system, but of course we don't have stdout so, I also don't have a solution.

... or maybe we could have it just be enabled for native builds, and have it be a no-op for embedded ones; that would be of _some_ help.
I don't recall though if it is easy to do, maybe something like `#![cfg_attr(not(feature = "python-bindings"), no_std)]` but for the logging definitions.

[...]

PR #283, opened 2024-05-28
Add logging

geonnave:
Towards #280. Still couldn't get `defmt-or-log` to work, but was able to have it working with just `log`.

Using `log` adds about 1 KB of overhead when using the no_std example (`cargo size --target='thumbv7em-none-eabihf' --no-default-features --features="crypto-cryptocell310, ead-authz, rtt" --release`)

This branch:
```
   text    data     bss     dec     hex filename
  66928       8    1120   68056   109d8 lakers-no_std
```

`main`:
```
   text    data     bss     dec     hex filename
  65852       0    1112   66964   10594 lakers-no_std
```

### DIFF
--- a/examples/coap/Cargo.toml
+++ b/examples/coap/Cargo.toml
@@ -19,3 +19,6 @@ coap-handler = "0.2"
 coap-handler-implementations = "0.5"
 coap-numbers = "0.2.3"
 coap-message-utils = "0.3.1"
+
+env_logger = "0.11.3"
+log = "0.4"
--- a/examples/coap/src/bin/coapclient.rs
+++ b/examples/coap/src/bin/coapclient.rs
@@ -2,6 +2,7 @@ use coap::CoAPClient;
 use coap_lite::ResponseType;
 use hexlit::hex;
 use lakers::*;
+use log::*;
 use std::time::Duration;
 
 const _ID_CRED_I: &[u8] = &hex!("a104412b");
@@ -16,6 +17,8 @@ const CRED_R: &[u8] = &hex!("A2026008A101A5010202410A2001215820BBC34960526EA4D32
 const _G_R: &[u8] = &hex!("bbc34960526ea4d32e940cad2a234148ddc21791a12afbcbac93622046dd44f0");
 
 fn main() {
+    env_logger::init();
+    info!("Starting EDHOC CoAP Client");
     match client_handshake() {
         Ok(_) => println!("Handshake completed"),
         Err(e) => panic!("Handshake failed with error: {:?}", e),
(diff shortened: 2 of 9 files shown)


chrysn:
Note that a user can still make this zero cost in the binary / at runtime by enabling the `log/release_max_level_off` feature, so I wouldn't worry about the cost here.

Do we need to defined policy around whether the security critical (hax compiled) parts may log? If critical parts don't log, that makes it easy on the safety/privacy aspects, because code outside that (even though being in the same crate) should probably not get access to secrets anyway. (We may want to vet debug impls for that though). If we log from critical code, is there a log level that is guaranteed not to contain key material? I'm unsure of the answer because while logged keys and secrets are useful during plug tests, they are a no-go in production systems. Not sure whether setting up cfg flags around it helps.

geonnave:
> make this zero cost by enabling the log/release_max_level_off feature

Thanks for the heads up, added a way to set it from the application.

As for the policy, yes that's important. As a first step, I was first looking into adding some sort of `skip` directive for the `Debug` derives in the state structs (e.g. to avoid logging secrets), but it turns out it is [still not supported in Rust](https://github.com/rust-lang/rust/issues/37009).

geonnave:
We can start with very basic logs just to "know where it went wrong", and build on top of it. Pushed a commit in that direction. Results look like this:

```
$ RUST_LOG=debug cargo run --bin coapserver
     Running `target/debug/coapserver`
[2024-05-29T07:09:37Z INFO  coapserver] Starting EDHOC CoAP Server
Waiting for CoAP messages...
Received message from 127.0.0.1:33736
[2024-05-29T07:09:40Z INFO  lakers] Initializing EdhocInitiator
[2024-05-29T07:09:40Z INFO  lakers] Enter process_message_1
[2024-05-29T07:09:40Z DEBUG lakers_shared::edhoc_parser] Enter parse_message_1
[2024-05-29T07:09:40Z DEBUG lakers_shared::edhoc_parser] Enter parse_suites_i
[2024-05-29T07:09:40Z INFO  lakers] Enter prepare_message_2
Received message from 127.0.0.1:36994
Received message 3
Found state with connection identifier ConnId(3)
[2024-05-29T07:09:40Z INFO  lakers] Enter parse_message_3
[2024-05-29T07:09:40Z DEBUG lakers_shared::edhoc_parser] Enter decode_plaintext_3
[2024-05-29T07:09:40Z INFO  lakers] Enter credential_check_or_fetch
[2024-05-29T07:09:40Z INFO  lakers] Enter verify_message_3
EDHOC exchange successfully completed
```

chrysn:
This level of "entering this function now" is usually called "tracing", and there is <https://docs.rs/log/latest/log/macro.trace.html> for that.

AIU non-application crates should not meddle with log features. If you still think we need the new feature, maybe document as:

"This has a global effect of disabling all logging through the log crate. Only set this from the application crate; it is a shortcut through adding a dependency to log and disabling all logging there."

geonnave:
@chrysn I changed the level to `trace!`, and moved the log dependency to the application, so that the library doesn't need an extra flag for that.

PR #284, opened 2024-06-04
Towards crypto & method agility: API updates, new buffer, new credential struct

geonnave:
Still very early stage.


geonnave:
So far this PR includes:
- updates in the API
  - support for setting different methods and crypto suites (even though only one is supported for now)
  - a separate function for setting identity (private key + credential)
- a new buffer struct that is parameterizeable via const generics
- a new credential struct that supports CCS's that hold asymmetric and symmetric keys
- improved handling of ID_CRED_x, including support for kids longer than 1 byte

Next steps are to update the examples and C/Python wrappers, then merge on main and continue the agility work on a new PR.

geonnave:
@malishav just asked for a review, mostly interested in an ack about the changeset (in comment above).

One point I have is that the Symmetric key type would not need to live in `main`, but not keeping it in `main` means having to keep another version of the code just for that, which in my opinion is not worth it.

geonnave:
Also, regarding actual crypto agility, I will leave for another PR since this one is already big enough...

malishav (approved):
I went briefly over the PR, it is massive, congrats on pulling this off! I am happy if you merge this!

Issue #289, opened 2024-06-23
lakers-python: Fails to build from dist tarball (at least on Python 3.13)


chrysn:
Where no wheels are available (on pypy as on #288, or on Python 3.13 because that's not released yet so doesn't get built by default I guess), builds fail related to the missing lakers-crypto-psa crate.

To reproduce:

```
$ docker run --rm -it docker.io/python:3.13-rc /bin/bash
$ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh /dev/stdin -y
$ . "$HOME/.cargo/env"
$ pip install cbor_diag
[succeeds -- just to show that Rust based crates can be built]
$ pip install lakers-python
[...]
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [23 lines of output]
      error: failed to load manifest for workspace member `/tmp/pip-install-galefl6n/lakers-python_3efcab44afde4189bc4835002ee9b6c9/lakers-python`
      referenced by workspace at `/tmp/pip-install-galefl6n/lakers-python_3efcab44afde4189bc4835002ee9b6c9/Cargo.toml`
      
      Caused by:
        failed to load manifest for dependency `lakers`
      
      Caused by:
        failed to load manifest for dependency `lakers-crypto`
      
      Caused by:
        failed to load manifest for dependency `lakers-crypto-psa`
      
      Caused by:
        failed to read `/tmp/pip-install-galefl6n/lakers-python_3efcab44afde4189bc4835002ee9b6c9/crypto/lakers-crypto-psa/Cargo.toml`
      
      Caused by:
        No such file or directory (os error 2)
      💥 maturin failed
        Caused by: Cargo metadata failed. Does your crate compile with `cargo build`?
        Caused by: `cargo metadata` exited with an error:
      Error running maturin: Command '['maturin', 'pep517', 'write-dist-info', '--metadata-directory', '/tmp/pip-modern-metadata-8f1fdte4', '--interpreter', '/usr/local/bin/python']' returned non-zero exit status 1.
      Checking for Rust toolchain....
      Running `maturin pep517 write-dist-info --metadata-directory /tmp/pip-modern-metadata-8f1fdte4 --interpreter /usr/local/bin/python`
      [end of output]
[...]
```

(Note that while pip complains about a newer version being available, `pip install --upgrade pip` doesn't change anything about the actual result)

geonnave:
Weird error about psa. 

[...]

PR #290, opened 2024-06-23
Make sdist packages usable

chrysn:
Closes: #289

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -5,6 +5,7 @@ on:
       branches: [main]
       tags: 'v*'
     workflow_dispatch:
+    pull_request:
 
 jobs:
   build-python-wheels:
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -48,10 +48,26 @@ jobs:
             maturin build --sdist --out wheelhouse
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
 
-      - name: Build wheels
+      # By building from sdist once we ensure that all components needed (see
+      # the dev-dependencies of lakers-python for an example) are available not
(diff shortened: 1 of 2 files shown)


chrysn:
I think this is ready for review; it passes CI.

Changes apart from "fixing the issue" are:
* On the platform that builds the sdist, also build the binaries from the sdist. (It'd be nice to build all from the sdist, but that'd create complex interactions, and chances are that that orthogonal niche doesn't attract bugs).
* CI runs Python tests now on every PR, helping spot similar issues early when more crates are added.

After this got merged, please consider adding the "build python wheels" check(s) to the mandatory list for merging.

geonnave:
> adding ... to mandatory list for merging

According to [this](https://docs.github.com/en/enterprise-server@3.11/actions/using-workflows/required-workflows#overview), I may not have permission to do that (seems to require organisation's `owner` permission). cc @malishav.

chrysn:
There's no hurry with that: having those builds visible and non-blocking would still be an improvement.

.github/workflows/python-wheels.yml
```diff
@@ -48,10 +49,21 @@ jobs:
             maturin build --sdist --out wheelhouse
         if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.11'
 
-      - name: Build wheels
+      # By building 
+      - name: Build wheels from sdist
```
geonnave, line 53: Here it is not obvious why we need to manually unpack and do the build from sdist (when building from source would normally just work).
From outer context, I know that it is probably to avoid things like https://github.com/openwsn-berkeley/lakers/issues/289
Can you add a comment on that?

chrysn: Oh, there's an incomplete comment anyway; fixed.

chrysn: Redid the comment once because the original version was not spot-on.

PR #292, opened 2024-06-24
Release: rust 0.6.2 and python 0.3.3

geonnave:
includes bug fixes with python wheel distribution
and in turn facilitates integration with libraries that depend on lakers-python, such as aiocoap

actually, the rust side doesn't have updates,
but since the triggers to publish python wheels are on github actions, I am just bumping all the versions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ default-members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.6.1"
+version = "0.6.2"
 repository = "https://github.com/openwsn-berkeley/lakers/"
 license = "BSD-3-Clause"
 readme = "shared/README.md"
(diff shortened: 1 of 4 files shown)


chrysn (approved):
LGTM, thanks a lot.

geonnave:
Seems to be something in the Actions environment, [tried to re-run 0.6.1](https://github.com/openwsn-berkeley/lakers/actions/runs/9626600783/job/26589723754) and got the same errors.

chrysn:
I've been trying to reproduce that locally, but that fails for lack of PSA, and GitHub runners are not exactly something to easily run locally.

This:

```
 error: expected one of `!` or `::`, found keyword `static`
 --> /home/runner/work/lakers/lakers/target/debug/build/psa-crypto-sys-3e14593d80be7147/out/shim_bindings.rs:3:11
  |
3 | rewriting static
  |           ^^^^^^ expected one of `!` or `::`
```

looks almost like a code generator started spewing debug output into the generated code. But those are generated by bindgen.

As presumably the tests run successfully locally, could you try checking in your Cargo.lock file? By now, that's [the default anyway](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control)

geonnave:
Yeah, I had `rustc 1.72.0-nightly` in my local machine (and it was working), then after updating to `rustc 1.81.0-nightly` I got the error.

chrysn:
I'm getting completely different errors on the PSA side; updating bindgen could help, trying that.

chrysn:
I've taken the liberty to push a commit onto your branch to run the CI of this PR again with the update from https://github.com/malishav/rust-psa-crypto/pull/1

Unless Mališa is currently on vacation or there's another reason that PR can't go in quickly, I'd rather avoid having that in the released version (so I can delete the branch again) -- but then again, it's a patched crate, and those are not part of what gets to crates.io anyway.

chrysn:
Hrmpf, that fixed my local build, but not the CI case.

At least I know now why it's on line 3: the top lines are `/* automatically generated by rust-bindgen 0.69.4 */` / empty; this means that `rewritingt static` can really no be part of valid Rust. Still no clue where that comes from precisely.

chrysn:
I can now reproduce locally: things worked on stable but broke on nightly.

I'm tempted to just fix this by editing out the "rewriting static" that's clearly a leftover println somewhere where I can't trace it. What I do know so far is that it does affect rust-psa-crypto.

chrysn:
The issue is tracked to bindgen: https://github.com/rust-lang/rust-bindgen/issues/2852

geonnave:
Wow, and that's a really fresh issue. Thanks for digging and finding it. Let's wait a bit and see if they fix it somewhat fast. Or else I can try to specify a previous rust version in the CI.

chrysn:
… but ultimately came from https://github.com/rust-lang/rustfmt/issues/6210. Rerunning things with a band-aid from https://github.com/malishav/rust-psa-crypto/pull/1 that disables rustfmt.

It is already fixed, but it'll take a day or two to get into the nightlies.

As a whole I think the best way to go is to revert my change (updating rust-psa-crypto's bindgen is still good but not critical for here), pin the nightly instead, and revert that pin after this is out.

chrysn:
All green now; the only change compared to your original branch is now the fixed nightly version, and that the example's nightly version is symlinked to the toolchain version to avoid missing one when changing back.

geonnave:
Great! Thanks for fixing that.

PR #294, opened 2024-07-01
Credentials also supporting symmetric keys

geonnave:
Work in progress. Leverages some ideas from #274. 

Built on top of #284, and supposed to be merged into the feature branch `crypto-method-agility`.


geonnave (commented):
@ElsaLopez133 thanks for pushing on parse_ccs_psk. Some comments below.

---
Please format the code with `cargo fmt` (or even better, install the pre-commit tool to do that automatically -- see the file .pre-commit-config.yaml).

---
>         // Why do they add +3 and +1 in CredentialPRK::parse

This comment can be sent over e.g. slack rather than in the code

---
Also please use a commit message following the pattern we use in the project (see previous commits to get a sense of it).

geonnave:
Not all tests pass yet, but this is now ready for review:
- new IdCred type that handles ID_CRED_x by value, by kid, etc., including encoding/decoding with the compact serialization when applicable
- new Credential type that supports EC2 and Symmetric keys (and thus can work with PSK, cc @ElsaLopez133), and plays along with the IdCred type
- fixes throughout the library to correctly compute MACs based on the right selected encoding of ID_CRED_x (before, we were always computing the MAC over an ID_CRED_x by reference, even when it was sent by value, as reported by @chrysn in #272).
- now the reference to a credential (e.g. a kid) can be a byte string, i.e. not limited to u8 anymore

In terms of code, something that still bothers me is the overuse of `map_err` in `cred.rs`, but I haven't yet found a better solution.

Also as stated in the first comment, this borrows ideas and code from #274.

lib/src/edhoc.rs
```diff
@@ -151,48 +147,24 @@ pub fn r_prepare_message_2(
...
-                        // just as in the CompactKid case the CredentialRPK is also incomplete.
-                        CredentialRPK {
-                            value: buffer,
-                            public_key: Default::default(),
-                            kid: Default::default(),
-                        }
-                    }
-                }
-            };
-
```
malishav, line 189: Could you comment on why we don't need this block any more?

shared/src/cred.rs
```diff
@@ -1,39 +1,187 @@
...
+            key: CredentialKey::EC2Compact(public_key),
+            kid: None,
+            cred_type: CredentialType::CCS,
+        }
+    }
+
+    /// Creates a new CCS credential with the given bytes and a pre-shared key
+    ///
+    /// This type of credential is to be used with the under-development EDHOC method PSK.
+    pub fn new_ccs_psk(bytes: BufferCred, symmetric_key: BytesKeyAES128) -> Self {
```
malishav, line 159: Do we want this in the current PR?

shared/src/cred.rs
```diff
@@ -55,40 +203,203 @@ impl CredentialRPK {
...
+                    key: CredentialKey::EC2Compact(public_key),
+                    kid: Some(BufferKid::new_from_slice(&[kid]).unwrap()),
+                    cred_type: CredentialType::CCS,
+                })
             } else {
                 Err(EDHOCError::ParsingError)
             }
         }
     }
+
```
malishav, line 225: I believe the PSK parsing routine should be added on top of the current PR as a separate branch, which we can merge once the PSK method becomes more stable

malishav (commented):
Congrats for pulling this one off! What remains to be done for this to be in a state where we can merge it? I left some comments inline, mainly for my understanding.

examples/lakers-no_std/src/main.rs
```diff
@@ -118,9 +123,16 @@ fn main() -> ! {
...
 
-        let (initiator, c_r, id_cred_r, ead_2) = initiator.parse_message_2(&message_2).unwrap();
+        let (mut initiator, _c_r, id_cred_r, _ead_2) =
+            initiator.parse_message_2(&message_2).unwrap();
         let valid_cred_r = credential_check_or_fetch(Some(cred_r), id_cred_r).unwrap();
-        let initiator = initiator.verify_message_2(I, cred_i, valid_cred_r).unwrap();
+        initiator
+            .set_identity(
+                I.try_into().expect("Wrong length of initiator private key"),
+                cred_i.clone(),
```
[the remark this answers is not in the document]
geonnave: It's because we need access to cred_i later in `credential_check_or_fetch(Some(cred_i), id_cred_i)`.

(but maybe this could be a reference, will take a look)

lib/src/edhoc.rs
```diff
@@ -109,28 +105,28 @@ pub fn r_prepare_message_2(
...
     );
 
-    let id_cred_r = match cred_transfer {
-        CredentialTransfer::ByValue => IdCred::FullCredential(cred_r.value.as_slice()),
-        CredentialTransfer::ByReference => IdCred::CompactKid(cred_r.kid),
-    };
-
     // compute ciphertext_2
-    let plaintext_2 = encode_plaintext_2(c_r, &id_cred_r, &mac_2, &ead_2)?;
+    let plaintext_2 = encode_plaintext_2(c_r, id_cred_r.as_encoded_value(), &mac_2, &ead_2)?;
```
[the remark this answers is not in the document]
geonnave: Pushed a commit with better documentation on both functions, hope it suffices?

lib/src/edhoc.rs
geonnave: According to [Section 5.4.3 of RFC 9528](https://www.rfc-editor.org/rfc/rfc9528.html#name-responder-processing-of-mess), ID_CRED_I is just made available to the application, which will process it. So even if it is a credential by value, we don't need to parse it here.

In addition, now the logic to parse ID_CRED_I is embedded in `decode_plaintext_3`, which calls `IdCred::from_encoded_value`. That's why we don't need to use different constructs for e.g. `CompactKid` or `FullCredential`.
Note that this only parses ID_CRED_I: from a (potentially compact-encoded) reference or value, e.g. `kid` / `{14: kccs}`, to the actual value of the ID_CRED_I, e.g.`{4: kid}` / `{14: kcss}`. Then, resolving the `kid` to a credential or verifying the syntax of the `kcss` is left to the application.

lib/src/edhoc.rs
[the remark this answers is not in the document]
malishav: Thanks!

lib/src/edhoc.rs
malishav: Sounds good, thanks for the explanation!

shared/src/cred.rs
geonnave: For now I documented it as experimental and moved the tests to a dedicated `test_experimental` module.

PR #295, opened 2024-07-05
Use defmt or log

WilliamTakeshi:
This PR addresses issue #280 . It introduces two new optional dependencies: `defmt` and `defmt-or-log`. Additionally, it makes the `log` dependency optional.

The purpose of this change is to allow users to specify either `log` or `defmt` as a feature for the lib. This flexibility should simplify debugging the lib on microcontrollers.

As this is my first PR for this library, I am open to feedback and suggestions! Please let me know if any changes are needed.

### DIFF
--- a/examples/coap/Cargo.toml
+++ b/examples/coap/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-lakers = { package = "lakers", path = "../../lib" }
+lakers = { package = "lakers", path = "../../lib", features = [ "log" ] }
 lakers-ead-authz = { path = "../../ead/lakers-ead-authz" }
 lakers-crypto = { path = "../../crypto/", features = [ "rustcrypto" ] }
 hexlit = "0.5.3"
--- a/examples/coap/Cargo.toml
+++ b/examples/coap/Cargo.toml
@@ -20,5 +20,13 @@ coap-handler-implementations = "0.5"
 coap-numbers = "0.2.3"
 coap-message-utils = "0.3.1"
 
+# Logging
 env_logger = "0.11.3"
-log = "0.4"
+defmt-or-log = { version = "0.2.1", default-features = false }
+log = { version = "0.4", optional = true }
+defmt = { version = "0.3", optional = true }
+
+[features]
+default = ["log"]
+defmt = ["dep:defmt"]
+log = ["dep:log"]
--- a/examples/coap/src/bin/coapclient.rs
+++ b/examples/coap/src/bin/coapclient.rs
@@ -1,8 +1,8 @@
 use coap::CoAPClient;
 use coap_lite::ResponseType;
+use defmt_or_log::info;
 use hexlit::hex;
 use lakers::*;
-use log::*;
 use std::time::Duration;
 
 const _ID_CRED_I: &[u8] = &hex!("a104412b");
--- a/examples/coap/src/bin/coapserver.rs
+++ b/examples/coap/src/bin/coapserver.rs
@@ -1,8 +1,8 @@
 use coap_lite::{CoapRequest, Packet, ResponseType};
+use defmt_or_log::info;
 use hexlit::hex;
 use lakers::*;
 use lakers_ead_authz::{ZeroTouchAuthenticator, ZeroTouchServer};
-use log::*;
 use std::net::UdpSocket;
 
 const ID_CRED_I: &[u8] = &hex!("a104412b");
--- a/examples/lakers-nrf52840/Cargo.toml
+++ b/examples/lakers-nrf52840/Cargo.toml
@@ -15,7 +15,7 @@ embassy-time = { git =  "https://github.com/embassy-rs/embassy", branch = "main"
 embassy-nrf = { git =  "https://github.com/embassy-rs/embassy", branch = "main", features = ["defmt", "nrf52840", "time-driver-rtc1", "gpiote", "unstable-pac", "time"] }
 
 # lakers
-lakers = { package = "lakers", path = "../../lib", default-features = false }
+lakers = { package = "lakers", path = "../../lib", features = [ "defmt" ] }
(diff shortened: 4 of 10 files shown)


geonnave:
Thanks for this PR @WilliamTakeshi! 

For my understanding:
- are we still able to use `RUST_LOG` to set the log level?
- since both `log` and `defmt` are optional, should I select one via features or there is one that is the default? (maybe `log` is the default since the CI just passed?)

Before merging, I would like to have it tested on an embedded device, like running the no_std example in the nRF52840. 
However, I am leaving for holidays next week and I don't have a board at home, so I will tag @malishav / @chrysn: if one of you guys could verify that it just works as before.

WilliamTakeshi:
> are we still able to use RUST_LOG to set the log level?

Yes, RUST_LOG is still supported (I have tested it using the `log` feature). For example, you can test it with the following commands:
```sh
RUST_LOG=error cargo run --bin coapclient
RUST_LOG=trace cargo run --bin coapserver
```
allowing you to see only logs on the server.


> since both log and defmt are optional, should I select one via features or there is one that is the default? (maybe log is the default since the CI just passed?)

By default, neither log nor defmt is enabled. You can enable either one by specifying the desired feature.  As for the CI passing, I wasn't aware that we had a test for the log, so it might have been luck that it passed!

In addition to specifying one of these features, you could also pass another logger configuration, such as `embassy-sync/defmt` or `embassy-sync/log`. However, it would be great if someone with an embedded device could test this to confirm.

malishav:
I can confirm I successfully tested the native examples with:

```
RUST_LOG=error cargo run --bin coapclient
RUST_LOG=trace cargo run --bin coapserver
```

That is OK.

I am not sure what exactly to test on nrf52840dk. I tried adding `defmt-or-log` and `defmt` as dependencies in `lakers-no_std` example which runs on nrf52840, and adding an info line in the source code. However, when I run this with `cargo embed` I get the `println!` prints in the output, but not the info line. I also tried setting the log level via `RUST_LOG` while running `cargo embed` but to no avail.

As a summary, it would be great if the PR included the changes for the `lakers-no_std` example and instructions how to test.

WilliamTakeshi:
@malishav Thanks for testing!

Unfortunately, I don't have an embedded device to test this on, making it harder for me to verify the implementation fully. Have you tried adding the features when importing `lakers`? For example:

for example: `lakers = { package = "lakers", path = "../../lib", features = [ "defmt" ] }` I also know that we can use others like:
* `"defmt-or-log/defmt"` 
* `"heapless/defmt-03"`
* `"embassy-sync/defmt"`
* `"embedded-hal-async/defmt-03"`

depending on which embedded device are you using.


On a side note, do you guys recommend any cheap device I can use to test it?

malishav:
> Have you tried adding the features when importing `lakers`? For example:
> 
> for example: `lakers = { package = "lakers", path = "../../lib", features = [ "defmt" ] }` I also know that we can use others like:
> 
> * ``

Yes, tried that, without it wouldn't build. It seems using `defmt` requires a special linker script, which is probably what is missing.

For embedded, our go-to hardware is the nRF52840 chip for now and we are extensively using [nRF52840-DK](https://www.nordicsemi.com/Products/Development-hardware/nRF52840-DK) development board.

WilliamTakeshi:
Sorry for the late reply!
I bought a nRF52840-DK  but it took a while to ship it to me.

@malishav, thank you for adding examples for the nRF52840! Using these examples, I was able to get `defmt_or_log` to work. Since I only have one board, I wasn't able to run the entire test, but I did start the initiator with the following command:

`PROBE_RS_PROBE=<PROBE_NUMBER> DEFMT_LOG=trace cargo run --bin initiator`

Here is the output I received:
```
INFO  Starting BLE radio
└─ initiator::____embassy_main_task::{async_fn#0} @ src/bin/initiator.rs:44  
INFO  init_handshake
└─ initiator::____embassy_main_task::{async_fn#0} @ src/bin/initiator.rs:59  
TRACE Initializing EdhocInitiator
└─ lakers::{impl#5}::new @ /home/william/projects/rust/lakers/lib/src/lib.rs:251 
TRACE Enter prepare_message_1
└─ lakers::{impl#5}::prepare_message_1 @ /home/william/projects/rust/lakers/lib/src/lib.rs:273 
TRACE Enter parse_message_2
└─ lakers::{impl#6}::parse_message_2 @ /home/william/projects/rust/lakers/lib/src/lib.rs:313
```

geonnave:
Thanks for the update @WilliamTakeshi, the output seems like what it should be, with the `trace` logs from the library showing up.
I wasn't able to make it work with logs in the no_std example, but that is using `cargo embed` which is deprecated, so I hope that once we update it to `probe-rs` the error goes away.

Can you add the following to the README before we merge?

<pre>
## Example: using logs
Logs can be used in both native and embedded applications. Once configured in an application, both can be controlled via environment variables:

- on native, set `RUST_LOG` to control Rust's built-in `log` facility
- on embedded, set `DEFMT_LOG` to control the [defmt](https://github.com/knurling-rs/defmt) crate

The selection of `log` or `defmt` is handled internally by the [defmt-or-log](https://github.com/t-moe/defmt-or-log) crate.

For example, `examples/lakers-nrf52840` is configured to use `defmt`. To globally enable logs at level `trace`:
```bash
DEFMT_LOG=trace cargo run --bin initiator
```

Different log levels can also be set per crate or module.
Here's how to globally set logs to level `trace`, while restricting `lakers` to level `info`:
```bash
DEFMT_LOG=trace,lakers=info cargo run --bin initiator
```
</pre>

WilliamTakeshi:
I also couldn't get `examples/lakers-no_std` to work. If needed, I can try updating it to use `probe-rs`.

By the way, should the new paragraph be added to the root `README.txt` or the `examples/lakers-nrf52840/README.txt`? I added on the root

geonnave:
>  I can try updating it to use probe-rs
That would be very welcome! Note that it's already tracked in #298.

And yes I think the right place is in the root.

Issue #298, opened 2024-07-17
Use probe-rs instead of cargo-embed


geonnave:
`cargo-embed` [has been deprecated](https://crates.io/crates/cargo-embed) and merged into `probe-rs`, so we should update the no_std example.

geonnave:
Solved by #299

PR #299, opened 2024-08-02
Use probe-rs instead of cargo embed

WilliamTakeshi:
I've taken a shot at removing cargo embed and using probe-rs instead. #298 

It's working using `cargo run --target="thumbv7em-none-eabihf"`

Additionally, I've switched to using `defmt` in place of `cortex_m_semihosting::hprintln`. If this change is undesirable, please let me know and I can revert it.

However, I encountered an issue: I couldn't get the tests to work with QEMU, as my computer gets stuck on every test. I tried running older commits, but nothing seems to work. Does `cargo run` with the default target work for you?

I'm marking this as a draft so we can resolve the QEMU issue before merging.

Thanks!


geonnave:
Thanks for this PR! 
Replacing `hprintln` with `defmt` is a good thing IMO. 
Do we still need (defmt-)`rtt` when we have `defmt`?
As for the QEMU issue, I will test it tomorrow morning (although `run-example-on-qemu` passed on CI, so I think it will be fine).

geonnave:
FYI, will test tomorrow instead since I cannot compile the psa backend in my M1 machine.

WilliamTakeshi:
If replacing `hprintln` with `defmt` is fine, I will update this PR from draft to ready for review!

> Do we still need `defmt-rtt` when we have `defmt`?

Yes, without `defmt-rtt` I was getting a bunch of linking issues like:
```
rust-lld: error: undefined symbol: _defmt_acquire
rust-lld: error: undefined symbol: _defmt_release
rust-lld: error: undefined symbol: _defmt_write
```
I found [a issue](https://github.com/knurling-rs/defmt/issues/132#issuecomment-680738096) that the solution was to keep the `use defmt_rtt as _;` to add a global logger


Let me know how the QEMU test goes!

geonnave:
I tried here, and I it indeeds gets stuck at 

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.83s
     Running `qemu-system-arm -cpu cortex-m3 -machine lm3s6965evb -nographic -semihosting-config enable=on,target=native -kernel /home/gfedrech/Developer/inria/dev/lakers-FORK/target/thumbv7m-none-eabi/debug/lakers-no_std`
Timer with period zero, disabling
```

On the other hand, the version on `main` prints:

```
$ cargo run  --no-default-features --features="crypto-psa" --release
    Finished `release` profile [optimized] target(s) in 4.36s
     Running `qemu-system-arm -cpu cortex-m3 -machine lm3s6965evb -nographic -semihosting-config enable=on,target=native -kernel /home/gfedrech/Developer/inria/dev/lakers-FORK/target/thumbv7m-none-eabi/release/lakers-no_std`
Timer with period zero, disabling
Hello, lakers!
Test test_new_initiator passed.
Test test_p256_keys passed.
Test test_prepare_message_1 passed.
Test test_handshake passed.
All tests passed.
```

(although I just realised that the QEMU version doesn't work without the `--release` flag, even on main).

---

Have you tried using `defmt-semihosting` as [mentioned here](https://defmt.ferrous-systems.com/setup)? It seems it should be used for testing with QEMU.

WilliamTakeshi:
@geonnave Thanks for the `defmt-semihosting` suggestion. I tried it out, but it appears that in addition to adding `defmt-semihosting`, I need to integrate a [printer](https://github.com/knurling-rs/defmt/blob/main/qemu-run/src/main.rs), a [decoder](https://github.com/knurling-rs/defmt/tree/main/decoder), and a [parser](https://github.com/knurling-rs/defmt/tree/main/parser) to properly process and display the data from QEMU.

This is how I understood they do in their example. Without this printer/decoder/parser combo, the log output appears as gibberish.

Because of this, I reverted to using `use cortex_m_semihosting::hprintln as info;` for the QEMU case and updated the README to reflect these changes.

Let me know if you need any further adjustments!

examples/lakers-no_std/README.md
```diff
@@ -16,10 +16,14 @@ be supported by QEMU's system emulation.
 This folder's `.cargo/config.toml` configures the target (`thumbv7m-none-eabi`)
 and some needed RUSTFLAGS, so this should pretty much just work(tm):
 
-    cargo run
+    cargo run --no-default-features --features="crypto-psa" --release
 
 You can exit QEMU pressing `CTRL-A`, then `X`. Or, if you're using tmux like
 me, `CTRL-A`, `A`, `X`.
 
+If you wanna try it on a embedded device, you can update the chip on `.cargo/config.toml` and run
```
geonnave, line 24: s/update the chip/modify the selected chip/

geonnave:
Thanks, left an editorial comment. Can you also please update the relevant section in the main README?

Feel free to add me as collaborator in your fork, this way I can also push some changes when needed (like updating the README).

examples/lakers-no_std/src/main.rs
```diff
@@ -6,15 +6,16 @@ use cortex_m_rt::entry;
 use cortex_m_semihosting::debug::{self, EXIT_SUCCESS};
 
 #[cfg(not(feature = "rtt"))]
```
geonnave, line 8: I think we should select based on the target, like this: 

```rust
#[cfg(target = "thumbv7m-none-eabi")]
```

examples/lakers-no_std/README.md
WilliamTakeshi: I think you can push changes on my branch.The default config is that:

> Allowing edits by maintainers
> If checked, users with write access to `openwsn-berkeley/lakers` can add new commits to your `remove_cargo-embed` branch. 


I did this change, but do you wanna try updating something else?

examples/lakers-no_std/src/main.rs
WilliamTakeshi: Looks like target doesn't work. [We don't have this option](https://doc.rust-lang.org/reference/conditional-compilation.html#set-configuration-options)

I used `#[cfg(target_abi = "eabihf")]` and `#[cfg(not(target_abi = "eabihf"))]` if this is fine

examples/lakers-no_std/README.md
geonnave: Oh, I think you are right, I should have tried :)

examples/lakers-no_std/src/main.rs
geonnave: Oh it's true. Looks good then!

examples/lakers-no_std/README.md
geonnave: In fact, for some reason it doesn't work :upside_down_face: 

```
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
To github.com:WilliamTakeshi/lakers.git
 ! [remote rejected] william/remove_cargo-embed -> william/remove_cargo-embed (permission denied)
error: failed to push some refs to 'github.com:WilliamTakeshi/lakers.git'
```

geonnave:
I also just noticed, could you rebase and use the `scope: description` pattern in the commit messages? Just to keep the same pattern as the rest of the commits. (basically just add the `example: ` prefix to all commits)

WilliamTakeshi:
> I also just noticed, could you rebase and use the `scope: description` pattern in the commit messages? Just to keep the same pattern as the rest of the commits. (basically just add the `example: ` prefix to all commits)

Sorry about that! I didn't knew you guys were using conventional commit messages. I've rebased the branch and updated the commit messages. Thanks for the heads up!

PR #319, opened 2024-11-03
python: Use Python's logging through pyo3_log

chrysn:
This makes logging accessible and configurable through Python.

### DIFF
--- a/lakers-python/src/lib.rs
+++ b/lakers-python/src/lib.rs
@@ -2,7 +2,6 @@
 /// Note that this module is not restricted by no_std.
 use lakers::*;
 // use lakers_ead_authz::consts::*;
-use env_logger;
 use lakers_crypto::{default_crypto, CryptoTrait};
 use log::trace;
 use pyo3::wrap_pyfunction;
@@ -72,8 +71,10 @@ impl AutoCredential {
 #[pyo3(name = "lakers")]
 fn lakers_python(_py: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
     // initialize the logger once when the module is imported
-    if env_logger::try_init().is_ok() {
-        trace!("lakers-python initialized from Rust side.");
+    if !pyo3_log::try_init().is_ok() {
+        // Not logging anything in the successful case as per pyo3_log recommendations: That would
+        // cache the current logging configuration, which likely changes soon after the imports.
+        log::error!("lakers-python failed to set up (different logger configured?)");
     }
 
     m.add_function(wrap_pyfunction!(p256_generate_key_pair, m)?)?;
--- a/lakers-python/src/lib.rs
+++ b/lakers-python/src/lib.rs
@@ -69,12 +69,15 @@ impl AutoCredential {
 // this name must match `lib.name` in `Cargo.toml`
 #[pymodule]
 #[pyo3(name = "lakers")]
-fn lakers_python(_py: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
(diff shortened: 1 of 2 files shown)


chrysn:
I've used this a lot with own added info messages; the trace messages emitted by lakers itself are not visible in this unless we touch [one more knob](https://docs.rs/pyo3-log/latest/pyo3_log/struct.Logger.html#method.filter) -- might make sense to turn that on by default, because nothing we do is performance critical on the Python side.

lakers-python/src/lib.rs
```diff
@@ -72,8 +71,10 @@ impl AutoCredential {
 #[pyo3(name = "lakers")]
 fn lakers_python(_py: Python, m: &Bound<'_, PyModule>) -> PyResult<()> {
     // initialize the logger once when the module is imported
```
geonnave, line 73:
```
// enables `logging.basicConfig(...)` from python and then logs from this wrapper just work
```

geonnave:
Cool! Suggested a comment just for clarity.

Could you enable that knob so that we don't lose the from-Rust log functionality that we currently have? Is it possible to have it controlled by an env var like with env_logger?

chrysn:
I'll add some API expositure and some basicConfig reference.

chrysn:
I misunderstood how this can be configured – more API won't do us any good. Instead, given that we're not logging a whole lot, and are not performance critical in Python, I've changed the code to unconditionally log trace from Rust. (This is still not super slow, as all pyo3_log's caching etc is still in). There are module level docs now that explain the logging isutation.

lakers-python/src/lib.rs
```diff
@@ -67,13 +66,26 @@ impl AutoCredential {
...
 
-// this name must match `lib.name` in `Cargo.toml`
+/// Lakers implementation of EDHOC.
+///
+/// The `EdhocInitiator` and `EdhocResponder` are entry points to this module.
+///
+/// Operations in this module produce logging entries on the `lakers.initiator` and
+/// `lakers.responder` logger names. Due to implementation details of `pyo3_log`, Python's log
+/// levels are cached in the Rust implementation. It is recommended that the full logging
+/// is configured before creating Lakers objects. A setup with `logging.basicConfig(loglevel=5)`
```
geonnave, line 76: Maybe best to add log level name? like `logging.basicConfig(level=logging.DEBUG)`

geonnave: Actually I am still confused on how to use it. 

I tried:

```python
import lakers, logging
logging.basicConfig(level=logging.DEBUG)
lakers.EdhocInitiator()
```

but after `python3 testlog.py` nothing got emitted.

So maybe I misunderstood the part about the hardcoded logs from Rust?

geonnave: I am also not sure it is a good idea to have Rust always-logging by default. I guess I could live with it sending only trace logs.
On the other hand, if we add more logs to the Rust side, we should make sure the python wrapper sets the Rust level to trace during build.

chrysn: So, the thing is that Python has no level below DEBUG (which Rust's log knows as 'trace'). That's why the docs say `basicConfig(level=5)`, which is even below DEBUG.

This also places it in a convenient spot where just enabling Python logging doesn't come on unintentionally, and pyo3_log's caching means that there should not be a practically relevant performance impact.

geonnave: Thank you, I just tested and it works as you said.

(previously I had misunderstood, and thought that traces logs would _always_ show even when python logs are off, but I verified that this is not the case)

geonnave:
Thanks for this PR! Merging

PR #312, opened 2024-10-24
Adding functions for message_4

ElsaLopez133:
Adding trait to call the edhoc_exporter and edhoc_prk_update either when in state in message 3 or 4.
Adding functions to prepare, parse, and process message 4.
Corrected examples (coap and lakers-nrf52840).
Added tests.

PR #321, opened 2024-11-04
python: Make safer to use

chrysn:
Previously, a call to `.edhoc_exporter()` would happily have produced exports from the all-zero key material.

This would have prevented the hiccup I ran into around https://github.com/openwsn-berkeley/lakers/pull/312#issuecomment-2454779410.

The present change ensures that attempting to run the EDHOC exporter results in a runtime error. aiocoap's tests still pass with it.

I'd prefer to go further, making the initiator start item also an option, or making this all into an enum, or even just using the high-level API, but this way at least we avoid inadvertently poking into data that should already be gone (which is something the high-level API would also ensure we don't do). Also I think that almost none of those items should implement Default. But let's do this here first and take it incrementally.

CC'ing @ElsaLopez133 (why can't I ask you to be a reviewer?) for review and because this might cause subtle (but easy to resolve) breakage with PR #312

### DIFF
--- a/lakers-python/src/initiator.rs
+++ b/lakers-python/src/initiator.rs
@@ -3,14 +3,19 @@ use lakers_crypto::{default_crypto, CryptoTrait};
 use log::trace;
 use pyo3::{prelude::*, types::PyBytes};
 
+use super::StateMismatch;
+
 #[pyclass(name = "EdhocInitiator")]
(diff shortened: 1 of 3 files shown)


geonnave (approved):
Went through and it looks good to me, thanks!

PR #325, opened 2024-11-25
IdCred enhancements

chrysn:
Some small fixes I found when updating Ariel OS to use the latest lakers:

* Some documentation on how IdCred is used would have helped, so I added an example that doubles as a doctest.
* Implementing that I found a typo (KCSS instead of KCCS); the fix is API compatible by deprecating the old version.
* The from_encoded_value constructor accepted some values that look like KIDs but are really longer. It may be a good idea to accept non-1-byte- KIDs, but this PR is designed for "all no-brainers so let's just merge it". I'm happy to provide longer KID support in there, but I don't know whether there's a limit on KID lengths in other places (won't hurt for this function to support more, though).
  [edit] Note that the way it is used internally (eg. decode_plaintext_3), the length is pre-checked to be correct due to `decoder.any_as_encoded()`, but that also means we're doing double work – still, a public function should do proper error handling.

### DIFF
--- a/shared/src/cred.rs
+++ b/shared/src/cred.rs
@@ -39,9 +39,18 @@ impl From<u8> for IdCredType {
     }
 }
 
-/// A value of ID_CRED_x: a credential identifier
+/// A value of ID_CRED_x: a credential identifier.
 ///
 /// Possible values include key IDs, credentials by value and others.
+///
+/// ```rust
+/// # use hexlit::hex;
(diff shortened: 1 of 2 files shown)


chrysn:
I think this is ready now; fixing things on cbindgen would have been excessive, so the comments in the doctest are in a less-than-ideal position as a workaround.

geonnave:
Looks good!

PR #327, opened 2024-11-27
parse_ccs fixes

chrysn:
parse_ccs is *very* lax right now, and that can cause great confusion: when sending a key in an unexpected format, a panic may happen deep inside p256 / subtle, around public keys being not a good point. (We should catch that panic either way, but this had sent me to the completely wrong track).

This adds a test for whether the parsing function rightfully errs out on stuff it doesn't recognize, and rewrites the parsing so that it *does* err out.

### DIFF
--- a/shared/src/cred.rs
+++ b/shared/src/cred.rs
@@ -217,50 +217,93 @@ impl Credential {
     /// If the given value matches the shape lakers expects of a CCS, i.e. credentials from RFC9529,
     /// its public key and key ID are extracted into a full credential.
     pub fn parse_ccs(value: &[u8]) -> Result<Self, EDHOCError> {
-        const CCS_PREFIX_LEN: usize = 3;
-        const CNF_AND_COSE_KEY_PREFIX_LEN: usize = 8;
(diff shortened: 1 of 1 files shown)


chrysn:
Two notes on the function's behavior, apart from the obvious new rejection of "malformed" items:
* It does accept longer KIDs now, simply because I'd have to go out of my way not to.
* It does not yet come down hard on invalid-length y coordinates, a) because they're not processed any further, but b) also because I cheated in earlier experiments and put y='' in to accommodate for credential-by-value space constraints. (On the long run, we should reject wrong y values, but maybe we should consider whether we regard y as optional, especially given that at least the credentials we're using now are only used for ECDH where the y doesn't matter).

chrysn:
The panicing is already tracked in https://github.com/openwsn-berkeley/lakers/issues/93, and there's a note in the code on it around https://github.com/openwsn-berkeley/lakers/blob/e1561d63f7a5e343a916852aae074524948e4384/crypto/lakers-crypto-rustcrypto/src/lib.rs#L118

PR #328, opened 2024-11-27
BytesKeyEC2: Prepare curve validation

chrysn:
Contributes-To: https://github.com/openwsn-berkeley/lakers/issues/93

A first step is to go from a type alias (which the enum `CredentialKey` can make no guarantees on, because pub enum variant members are always pub) into a newtype.

This is a breaking API change, so maybe we should prioritize doing the API change even when we can't validate, but this will need a bit more than this single commit.

### DIFF
--- a/shared/src/cred.rs
+++ b/shared/src/cred.rs
@@ -6,12 +6,12 @@ pub type BufferIdCred = EdhocBuffer<192>; // variable size, can contain either t
 pub type BytesKeyAES128 = [u8; 16];
 
 #[derive(Debug, Copy, Clone, PartialEq)]
-pub struct BytesKeyEC2([u8; 32]);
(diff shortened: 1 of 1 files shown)


chrysn:
Open points:

* I'd consider shedding a few derived traits:
  * PartialEq is necessary because CredentialKey has it too. Is this relevant outside of testing? (Doesn't hurt too much, either).
  * I'm conflicted about the try-from-slice -- it's convenient transitionally because most creation happened from slices anyway, but … dunno.

* How do we do the verification? When the type is created, that's typically far down in the parsing step, and we don't have access to a Cryptography.

chrysn:
More precise open question: Should we even be doing this on BytesKeyEC2, or should that not be BytesP256ElemLen that would become a coordinate type?

Issue #330, opened 2025-01-16
MAX_MESSAGE_SIZE configuration granularity is insufficient


chrysn:
There's a note in shared:

```rust
// TODO: find a way to configure the buffer size
```

The current workaround is to have a feature `quadruple_sizes` that scales *everything* up; turns out that for at least one application, 4x is so large it smashes the stack, and 1x is too small for ACE tokens.

I've played around with envinroment variables, and found something that works:

```patch
-// TODO: find a way to configure the buffer size
-// need 128 to handle EAD fields, and 192 for the EAD_1 voucher
-pub const MAX_MESSAGE_SIZE_LEN: usize = SCALE_FACTOR * (128 + 64);
+pub const MAX_MESSAGE_SIZE_LEN: usize = if let Some(e) = option_env!("LAKERS_MAX_MESSAGE_SIZE") {
+    match konst::primitive::parse_usize(e) {
+        Ok(n) => n,
+        Err(_) => panic!("Error parsing LAKERS_MAX_MESSAGE_SIZE"),
+    }
+} else {
+    // need 128 to handle EAD fields, and 192 for the EAD_1 voucher
+    SCALE_FACTOR * (128 + 64)
+};
```

But I'm not happy with that: It requires the environment variable to be set from outside, and that means that the crate that has the requirement somewhere down in the tree needs to parse the environment variable as well, and complain, and then the user has to call it with some variable … nah.

I do now think that the better form is to have granular `lakers_max_message_size_256` etc. features, in arbitrary steps, which can be selected by downstream crates; that seems to be the common practice in [other embedded crates as well](https://docs.rs/embassy-executor/latest/embassy_executor/#task-arena-size).

A prettier solution is of course to introduce generics, but that'll need more changes, and I'd rather have something working soon.

chrysn:
Two things I've added:

* Features for other buffer sizes I ran into
* I've moved the branch onto 0.7.2. It works with main too, so merging it is fine, but to avoid working from different APIs, I need it to be usable before the message 4 API change was merged.

geonnave:
Thanks for raising this and referencing `embassy-executor`. This approach seems fine to me.

PR #331, opened 2025-01-16
shared: Introduce max_message_size_… features

chrysn:
Closes: https://github.com/openwsn-berkeley/lakers/issues/330


shared/src/lib.rs
```diff
@@ -42,9 +42,20 @@ pub const SCALE_FACTOR: usize = 1;
 #[doc(hidden)]
 pub const SCALE_FACTOR: usize = 4;
```
geonnave, line 43: Is the reason for keeping `SCALE_FACTOR`, despite the additions of the new features, that 
  it allows quick experimentation with larger buffer sizes if needed?

geonnave:
Thank you for the PR. Looking into why CI fails.

shared/src/lib.rs
chrysn: If we're already in a breaking change, we could replace scale_factor with a max_buffers feature.

chrysn: Adding this; turns out only 2 more places use SCALE_FACTOR (and one of them was completely unused).

I'll add a "very big" variety for the unconstrained systems to replace scale_factor.

chrysn: SCALE_FACTOR is now gone in updates; there is the option to scale everything to 1KiB (which is a sensible maximum at least in CoAP, where doing more would mean fragmenting even on a common IPv6 transport); the quadruple_sizes feature was replaced with a large_buffers feature.

geonnave: Looks great!

chrysn:
Unlike on #333 I did *not* rebase this after #339 to fix CI, but instead merged main back: This enables this PR to stay on 0.7.2,  i.e. to be a single fixing change on the latest release (which is how I use it on Ariel OS).

@geonnave, is there anything more you'd like to review? Can we merge this as is, or should we trick around (possibly by briefly disabling CI checks) to preserve both a "feature branch simply merges back" shape and the "this can be picked into a 0.7 branch" property? (I think it's fine to merge).

chrysn:
Aaaargh: cbindgen doesn't support switching by cfg (while it supported the old idiom of cfg'd items).

Emulating the old idiom doesn't scale well (the entry for the shortest size would need to be not(any(feature = "first", feature = "second", ...))), I'll look for alternatives.

(ceterum censeo: exposing the API to C is a liability for this project)

geonnave:
Merging!!

PR #333, opened 2025-01-22
Change API for EAD items to support large numbers

chrysn:
The set of supported values is not changed, but the API is changed so that the implementation can be fixed without an extra breaking change.

---

Didn't run any special tests, let's see if CI is happy. (Should be a no-brainer though).

### DIFF
--- a/ead/lakers-ead-authz/src/lib.rs
+++ b/ead/lakers-ead-authz/src/lib.rs
@@ -12,7 +12,7 @@ pub use device::{ZeroTouchDevice, ZeroTouchDeviceDone, ZeroTouchDeviceWaitEAD2};
 pub use server::{ZeroTouchServer, ZeroTouchServerUserAcl};
 
 pub mod consts {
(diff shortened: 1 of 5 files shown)


chrysn:
I had a hunch I'd done this before already: this was my first PR (https://github.com/openwsn-berkeley/lakers/pull/100/files). But too much has changed, so the old PR is not actionable any more.

It did more than just change the type (it made its details private); that's still something worth doing, but maybe better deferred to be done together with other restructurings like supporting multiple EAD items.

chrysn:
Rebased onto main to fix CI; I think this is ready for review.

chrysn:
Latest fix also adjusts for the language bindings that did not get tested as part of `cargo test`; relying on CI to check whether those are still good.

geonnave:
Almost ready to merge, can you add `ead: ` as prefix to the commit messages?

chrysn:
Fixed!

PR #339, opened 2025-02-02
CI: Install cbindgen

chrysn:
On top of https://github.com/openwsn-berkeley/lakers/pull/337, this installs cbindgen, which is apparently still missing.


chrysn:
@malishav, can you merge https://github.com/malishav/rust-psa-crypto/pull/1? Then I can rebase that and drop the #337 fix, and we got CI back.

chrysn:
Removed the commit that switches to my branch.

I don't know how anything would go wrong now, but still, let's see what CI says. If all is green, we have unblocked the other issues; I think we'll have to rebase all pending PRs once this is merged.

Issue #336, opened 2025-01-29
feature: Directly ingest naked COSE key


chrysn:
The `parse_ccs()` Credential constructor takes a full CCS. [RFC9528 Section 3.5.2](https://datatracker.ietf.org/doc/html/rfc9528#section-3.5.2) also describes that naked COSE keys can be used (and I think are common in ACE EDHOC profile) that are turned into a CCS by prefixing them with A108A101.

I suggest (and will probably later PR) to add a `dress_cosekey()` constructor that takes only a COSE key, extracts the key (possibly refactoring with `parse_ccs`) and adds the prefix. This is a minor usability help, but saves downstream crates from carrying a 4-byte magic byte sequence around.

Ideally, that parsing function should be less strict than what `parse_ccs` does now; in particular, it should support EC2 keys where only the X coordinate is given (as is sufficient for ECDH), and which have no kid is present. Is there any particular reason why this (and a subject) is required in `parse_ccs()`, other than that `parse_ccs()` was originally designed to ingest test vectors where those happened to be present?

PR #341, opened 2025-02-03
Add parser/dresser for naked COSE keys

chrysn:
Implements https://github.com/openwsn-berkeley/lakers/issues/336, and refactors the `parse_ccs()` function, broadening the range of accepted items.

Closes: #336

### DIFF
--- a/shared/src/cred.rs
+++ b/shared/src/cred.rs
@@ -359,6 +359,103 @@ impl Credential {
         }
     }
 
+    /// Parse a COSE Key, accepting only understood fields.
+    ///
+    /// This takes a decoder rather than a slice because this enables a naked decoder to assert that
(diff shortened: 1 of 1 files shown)


shared/src/cred.rs
```diff
@@ -359,6 +312,103 @@ impl Credential {
         }
     }
 
+    /// Parse a COSE Key, accepting only understood fields.
+    ///
+    /// This takes a decoder rather than a slice because this enables a naked decoder to asset that
```
geonnave, line 317: to assert* ?

geonnave:
Looks good to me, left just a minor comment, then I am ready to merge.

shared/src/cred.rs
chrysn: Thanks, fixed.

PR #343, opened 2025-02-05
shared/cred: Allow absent subject

chrysn:
https://github.com/openwsn-berkeley/lakers/pull/341 had an easy follow-up -- was there ever a reason to require a CCS to have a subject, other than that it is there in the examples?

With this, I think that the parser can deal with all the credentials which I've previously created manually.

---

The loop expression removes the assurances on the order, but the processor already does not enforce full deterministic encoding.


geonnave:
Thanks, looks good to me.

PR #349, opened 2025-03-18
Add cargo check on CI

WilliamTakeshi:
This PR introduces `cargo check`.

My plan is to gradually increase the coverage over multiple PRs and eventually add Clippy. Through these changes, I'm also familiarizing myself with the codebase.

* [fix: cargo check unexpected_cfgs](https://github.com/openwsn-berkeley/lakers/commit/c798e11bcbcdfb57adaedc2b2833c8d32c30ebd0) remove unnused cfgs (https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html)
* [fix: dead_code and unnecessary mut](https://github.com/openwsn-berkeley/lakers/commit/8bf1f8cc3086aa55485ab700bcdb794924e46444) add `_` on unused vars and remove unnecessary `mut`s
* [chore: allow deprecated for from_int_raw](https://github.com/openwsn-berkeley/lakers/commit/aa6ad122dd7ee396cba0745c8a4d88adbf173e8e) I'm marking fine the use of some deprecated functions for now, until we have a easy replacement


.github/workflows/build-and-test.yml
```diff
@@ -18,6 +18,8 @@ jobs:
       uses: actions/checkout@v4
     - name: Install Rust formatter
       run: rustup component add rustfmt
+    - name: Run cargo check
+      run: cargo check
```
chrysn, line 22: I think this should be `--workspace`, and eventually fail (and not just show, nobody reads CI logs) if something is spotted.

Currently, wouldn't go through due to lakers-c apparently being embedded-only, but that should be fixable.

chrysn (commented):
That looks like a good change as a whole.

.github/workflows/build-and-test.yml
chrysn: … and maybe with a second --all-features pass, but that too currently fails :-/

geonnave:
Thanks for this PR, improving warnings and cleaning the codebase is very welcome. 

I just saw emails that some of my github tokens expired. I just updated them. I hope the CI will be ok!

(I am in the middle of some weeks of mission abroad / PTO, doing some catch up now!)

.github/workflows/build-and-test.yml
WilliamTakeshi: Hey @chrysn , thanks for the feedback!

I’ve added `RUSTFLAGS=-Dwarnings` to the CI so that any warnings will now cause the build to fail.

I also added the `--all-features` flag to the check. The main change is in how the `Crypto` type is selected. Previously, the features "psa", "rustcrypto", and "cryptocell310" were mutually exclusive. I’ve switched to using `cfg_if!` to make them additive instead — though this introduces an implicit priority order. Let me know if the project prefers to keep them mutually exclusive instead.

I plan to add `--workspace` in the future, but for now I’m holding off as I’m not yet confident making changes to that part.

geonnave: For all of our use cases so far, using a single backend has been enough, i.e., mutually exclusive is fine. 
There might be use cases for having more than one backend used at once (benchmarking? multi-platform code?), but I am not sure. In any case, as long as we don't lose anything in terms of the current behavior, it should be fine.

geonnave:
I will just test this in my nRF52840 tomorrow morning and then merge it.

PR #350, opened 2025-03-22
python: Enhance error handling

chrysn:
I just spent way more time than I'm happy to admit debugging a case of feeding in the wrong kind of credential.

This PR goes through one commit of simplifying Python side error handling (no point doing the match to avoid the early returns when code is not hax'd), and then goes about adding some hints as to *what* failed to errors.

### DIFF
--- a/lakers-python/src/initiator.rs
+++ b/lakers-python/src/initiator.rs
@@ -181,7 +181,7 @@ impl PyEdhocInitiator {
     /// Key material can be extracted after this point, but some properties of the protocol only
     /// hold when non-EDHOC messages protected with the extracted key material are received from
     /// the peer.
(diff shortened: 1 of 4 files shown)


chrysn:
The StateMismatch errors just weren't pretty. This adds a bit more work, but Python is aimed to be easy to use, so I think some runtime code to assess the state of affairs so that a good error message can be produced is worth it.

Also, this now brings the tools to have a sensible `__repr__`, which also helps putting the errors in context.

geonnave:
Went through and looks good! 

Didn't have the chance to test the results, but I see the unit tests, which look good. 

Ready to merge once conflicts are solved.

geonnave:
Just ping-ing here, happy to merge after conflicts are solved :)

Issue #347, opened 2025-03-16
C error len is not set


geonnave:
The len field in EdhocMessageBuffer is not being set.

I isolated this to happen just after merge of PR #333.

Logs:

```
~/D/i/d/l/e/lakers-c-native ❯❯❯ make clean all LAKERS_EAD=authz && ./lakers_c_native                                                                                    ✘ 1 tags/v0.8.0~4 ✭ ◼
rm -f main.o lakers_c_native
gcc -Wall -I../../target/include -fsanitize=address,undefined,leak -DLAKERS_EAD_AUTHZ -ggdb -c main.c
gcc -Wall -I../../target/include -fsanitize=address,undefined,leak -DLAKERS_EAD_AUTHZ -ggdb -O0 -o lakers_c_native main.o -L../../target/x86_64-unknown-linux-gnu/release/ -llakers_c -lcoap-3
Calling lakers from C!
loading credentials.
creating edhoc initiator.
creating ead-authz device.
computing authz_secret.
computing ead_1.
0 bytes: 
Begin test: edhoc initiator.
0 bytes: 
sending msg1
sending coap message of size 0+1
1 bytes: F5
received coap response
1 bytes: F5
processing msg2
Error parse msg2: 6
```

geonnave:
Still confused about why it happens, `label` was correctly updated to `u16`

```
pub struct EADItemC {
    pub label: u16,
    pub is_critical: bool,
    pub value: EdhocMessageBuffer,
}
```

malishav:
Just bumping this up as it bit me today

chrysn:
To reproduce this, I went through the following steps:
* unset the global Cargo config target-dir (already for the C building, where I had to ignore several warnings)
* switch from -lcoap-3 to -lcoap-3-gnutls (working off libcoap develop branch, README is not clear on versions)

and because I didn't install libcoap system-wide:
* s/=/+=/ on the CFLAGS and LDFLAGS in Makefile
* set `CFLAGS="-I ~/git/libcoap/include"` in the environment
* explicitly add LDFLAGS in the TARGET rule
* run as CFLAGS="-I ${HOME}/git/libcoap/include" LDFLAGS="-L ${HOME}/git/libcoap/.libs/" make clean all LAKERS_EAD=authz && LD_LIBRARY_PATH=${HOME}/git/libcoap/.libs/ ./lakers_c_native

… but all that still only tries to run a CoAP server, expecting a peer that is not documented in the C test's README.

[...]

PR #353, opened 2025-03-27
c: Align manual ifdefs with feature-to-constant #[cfg]s

chrysn:
This fixes mistakes introduced in #331.

Closes: https://github.com/openwsn-berkeley/lakers/issues/347


malishav:
Just tested it in my setup, it works! Thanks for providing this PR!

PR #354, opened 2025-03-27
Fixes to C example

chrysn:
The example has grown outdated, and its Makefile was not extensible.

Found while debugging https://github.com/openwsn-berkeley/lakers/issues/347

This should be checked by CI, but until it is, this patches current breakage.


geonnave:
Looks good, thanks!

PR #351, opened 2025-03-23
python: Add documentation

chrysn:
This adds all the steps that once the package is installed, `sphinx-build doc html` in the lakers-python directory will produce reasonably usable HTML docs. Those could be autogenerated on readthedocs (I can set that up) or on hosted pages if you want to play with GitHub (I prefer the former).

Missing steps:

* [ ] Find which workaround for https://github.com/PyO3/pyo3/issues/4326 we want to use. In particular, this affects `Credential`, but I'm a bit reluctant to start fixing it there because the documentation is outdated from the back-and-forth on non-CCS credentials.
* [x] Get an account at readthedocs and publish there
* [ ] Do we wait for https://github.com/openwsn-berkeley/lakers/issues/282, or do nothing, or manually add some more hints so that users get any idea on which types to throw in where?

Closes: https://github.com/openwsn-berkeley/lakers/issues/345

PR #355, opened 2025-03-27
python: Provide documentation URI

chrysn:
I'm not 100% sure all the settings are right, but this does produce a PKG-INFO in the source tar ball that has a `Project-URL: Documentation, …` added, and has a description. ([Docs](https://www.maturin.rs/metadata) indicate that adding a dynamic on project_url should also give the right results, but I'm not sure what we'd have to set for that).

Follow-up-for: https://github.com/openwsn-berkeley/lakers/pull/351

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ resolver = "2"
 
 [workspace.package]
 version = "0.8.0"
-repository = "https://github.com/openwsn-berkeley/lakers/"
+repository = "https://github.com/lake-rs/lakers/"
 license = "BSD-3-Clause"
 readme = "shared/README.md"
 keywords = ["iot", "security", "protocol", "crypto", "edhoc"]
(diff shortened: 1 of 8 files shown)


geonnave:
Just to clarify: What do you mean by "not 100% sure all the settings are right"? Do you need some input here?

chrysn:
I mean that I'm unsure whether project name and and version=['dynamic'] *should* really be set this way, and whether the free-form keys in project.urls wind up in the right place in PyPI.

I think they are, and there's justification from tools complaining if they're not set, so … unless we want to spend lots of time understanding the How and Why and maybe some history of what the Python Packaging Authority has stated over time, it's probably Good Enough to try this, and enhance gradually if anything about the presentation of the next release is off. (Gut feeling is that even if it's broken, this will rather be an improvement given [the index](https://pypi.org/search/?q=lakers%2Dpython) currently says "lakers-python: None" and there are barely any links out).

Issue #352, opened 2025-03-23
Tolerate GREASE


chrysn:
Tracking issue -- to work with any peer, we'll need

* [ ] Support for the full range of EAD numbers (already in the API)
* [ ] Support for multiple EAD items (API change)

geonnave:
Cool, this is also aligned with supporting the combined operation of lake-authz and lake-ra, see https://github.com/lake-wg/ra/issues/12, cc @ysong02

PR #356, opened 2025-05-01
Eaditem option to array

WilliamTakeshi:
**Request for Comments**

This PR is an initial step toward addressing issue #177. I'm experimenting with changing the internal representation of `EADItem` from an `Option` to a fixed-size array (`[EADItem; MAX_EAD_ITEMS]`). I chose 4 as a tentative maximum number of items, but that’s open for discussion.

I’m opening this early as a Request for Comments (RFC) to get feedback from folks with more context. In particular:

    This introduces significant breaking changes, replacing `Option<EADItem>` with `[EADItem; MAX_EAD_ITEMS]`.
    → Is it acceptable to move forward with this breaking change?
    → Or would it be better to maintain backward compatibility by duplicating the relevant APIs and deprecating the old ones?

Additionally, I could use some help with the C library integration — I wasn’t able to get it working yet (skill issues :cry: ).

Any thoughts, suggestions, or guidance would be greatly appreciated. Thanks in advance!

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -63,6 +63,7 @@ pub const MAC_LENGTH: usize = 8; // used for EAD Zeroconf
 pub const MAC_LENGTH_2: usize = MAC_LENGTH;
 pub const MAC_LENGTH_3: usize = MAC_LENGTH_2;
 pub const ENCODED_VOUCHER_LEN: usize = 1 + MAC_LENGTH; // 1 byte for the length of the bstr-encoded voucher
+pub const MAX_EAD_ITEMS: usize = 4;
 
 // maximum supported length of connection identifier for R
 //
@@ -488,7 +489,7 @@ pub struct ProcessingM2 {
     pub plaintext_2: EdhocMessageBuffer,
     pub c_r: ConnId,
     pub id_cred_r: IdCred,
-    pub ead_2: Option<EADItem>,
+    pub ead_2: [EADItem; MAX_EAD_ITEMS],
 }
 
 #[derive(Default, Debug)]
@@ -507,7 +508,7 @@ pub struct ProcessingM3 {
     pub th_3: BytesHashLen,
     pub id_cred_i: IdCred,
     pub plaintext_3: EdhocMessageBuffer,
-    pub ead_3: Option<EADItem>,
+    pub ead_3: [EADItem; MAX_EAD_ITEMS],
 }
 
 #[derive(Default, Debug)]
@@ -676,7 +677,7 @@ impl TryInto<EdhocMessageBuffer> for &[u8] {
 }
 
 #[cfg_attr(feature = "python-bindings", pyclass)]
-#[derive(Clone, Debug)]
+#[derive(Clone, Default, Debug)]
 pub struct EADItem {
     /// EAD label of the item
     ///
@@ -696,6 +697,10 @@ impl EADItem {
             is_critical: false,
             value: None,
         }
+    }
+
+    pub fn new_array() -> [EADItem; MAX_EAD_ITEMS] {
+        core::array::from_fn(|_| EADItem::new())
     }
 }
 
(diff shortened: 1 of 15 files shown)


geonnave:
Very very interesting, thanks for this PR! From a quick glance it looks in the right direction. Will give a more detailed look as soon as possible and let you know.

chrysn:
This would also addres some of https://github.com/openwsn-berkeley/lakers/issues/352.

Having a breaking change is fine in this situation I think.

geonnave:
@WilliamTakeshi  and as for lakers-c, a first step would be to make it just compile as is today (without extending the C API to support multiple EAD items). It would be something like this:

```patch
diff --git a/lakers-c/src/initiator.rs b/lakers-c/src/initiator.rs
index 1ce1cb4..13fa1b3 100644
--- a/lakers-c/src/initiator.rs
+++ b/lakers-c/src/initiator.rs
@@ -60,16 +60,15 @@ pub unsafe extern "C" fn initiator_prepare_message_1(
         ConnId::from_int_raw(*c_i)
     };

-    let ead_1 = if ead_1_c.is_null() {
-        None
-    } else {
+    let mut eads_1 = EADItem::new_array();
+    if !ead_1_c.is_null() {
         let ead_1 = (*ead_1_c).to_rust();
-        Some(ead_1)
-    };
+        eads_1[0] = ead_1;
+    }

     let state = core::ptr::read(&(*initiator_c).start);

-    let result = match i_prepare_message_1(&state, crypto, c_i, &ead_1) {
+    let result = match i_prepare_message_1(&state, crypto, c_i, &eads_1) {
         Ok((state, msg_1)) => {
             core::ptr::write(&mut *message_1, msg_1);
             core::ptr::write(&mut (*initiator_c).wait_m2, state);
```

To compile, `cd lakers-c` and run `./build.sh crypto-rustcrypto`.
This will try to build the C library for the host architecture (your computer).

After the changes (similar to the one in the patch above), eventually it will compile. 
Then you can `cd examples/lakers-c-native` and `make LAKERS_EAD=authz && ./lakers_c_native`. This will expect the coapserver to be running in another terminal (`cargo run --bin coapserver`).

malishav:
>Having a breaking change is fine in this situation I think.

Agreed!

shared/src/lib.rs
```diff
@@ -582,7 +583,7 @@ pub type EdhocMessageBuffer = EdhocBuffer<MAX_MESSAGE_SIZE_LEN>;
 
 /// An owned EAD item.
 #[cfg_attr(feature = "python-bindings", pyclass)]
-#[derive(Clone, Debug)]
+#[derive(Clone, Default, Debug)]
```
chrysn, line 586: In a [parallel PR](https://github.com/openwsn-berkeley/lakers/pull/364) I've removed Default from places where there is no sensible default -- why is there a Default here? (The derived Default is ... a single byte of padding? Is that a sensible default, and does this library even support sending padding?).

chrysn:
There was a brief (probably merge-test?) PR #375 on this recently.

How can we best make progress on this? So far I hope they are trivial to resolve -- but I'd like to continue work on enhancing the way we use buffers, and those would not be without major conflict, so I'd hope to get this PR in soon.

shared/src/lib.rs
WilliamTakeshi: Removed the `Default`!

Looks like I didn't needed it

shared/src/lib.rs
```diff
@@ -508,7 +509,7 @@ pub struct ProcessingM2 {
     pub plaintext_2: BufferPlaintext2,
     pub c_r: ConnId,
     pub id_cred_r: IdCred,
-    pub ead_2: Option<EADItem>,
+    pub ead_2: [EADItem; MAX_EAD_ITEMS],
```
chrysn, line 512: IIUC, this regards EADItem as internally null-able -- an item with `.value.is_none()` is "not really present" and will be skipped.

This is not quite how EAD items should work per spec: ead_value is optional ([3.8](https://datatracker.ietf.org/doc/html/rfc9528#name-external-authorization-data)), and there is no indication that the absent ead_value is in any way equivalent to the empty byte string.

An easy save is probably `[Option<EADItem>; MAX_EAD_ITEMS]`, especially since `MAX_EAD_ITEMS` is small, and `EADItem` has a niche, so the option won't make it really bigger. It might make sense to give that an `EadItems` type even, to hide the implementation details.

shared/src/lib.rs
```diff
@@ -651,41 +656,102 @@ mod helpers {
...
+                        u32::from_be_bytes(len_bytes.try_into().unwrap()) as usize
+                    }
+                    27 => {
+                        let len_bytes = input
+                            .get(offset..offset + 8)
+                            .ok_or(EDHOCError::ParsingError)?;
+                        offset += 8;
+                        let len_u64 = u64::from_be_bytes(len_bytes.try_into().unwrap());
+                        if len_u64 > usize::MAX as u64 {
+                            return Err(EDHOCError::ParsingError);
```
chrysn, line 726: A funny thing here is that at some point (not necessarily in this PR, but more widely as per https://github.com/openwsn-berkeley/lakers/issues/367) we'll want to check if deterministic CBOR was used -- which would add a clause `if this is <= u32::MAX, return Parsing Error` -- in which case a good compiler should deduce that the whole case of 27 unconditionally produces a ParsingError on 32bit systems, and can thus be folded in to the `_` case :-)

shared/src/python_bindings.rs
```diff
@@ -58,6 +58,11 @@ impl EADItem {
     fn is_critical(&self) -> bool {
         self.is_critical
     }
+
+    #[staticmethod]
+    fn new_empty_array_py() -> [Self; MAX_EAD_ITEMS] {
```
chrysn, line 63: I don't think that `[EADItem; MAX_EAD_ITEMS]` is a good idiomatic Python version; those should take an iterator and produce a Python array with no nulled elements. But if you don't have any special interest in Python, it's probably fine this way (make-it-work) and I can continue with a make-it-pretty followup.

chrysn (commented):
I think this is on the right track; there's a critical concern about value-less EAD items though (see comments). Addressing that with a newtype should also help keep our API flexible: On the long run I'd like to get rid of the `pub` to get some flexibility in changing internals without breaking the API.

shared/src/lib.rs
WilliamTakeshi: I really liked the idea of using a newtype — it gives us flexibility to change the struct later without breaking anything. I put together a commit here, WDYT?? [`33e9cdc` (#356)](https://github.com/openwsn-berkeley/lakers/pull/356/commits/33e9cdcb0396d9359ea7ff1f83411f59a4409398)

shared/src/lib.rs
WilliamTakeshi: Removed here! [`c36c838` (#356)](https://github.com/openwsn-berkeley/lakers/pull/356/commits/c36c838c4ea4179e469d907c1e36ab77b9daaf1d)

shared/src/python_bindings.rs
WilliamTakeshi: humm, I don't know what you mean by an iterator. But using the newtype I think we can expose the creating by using our methods instead of letting the user create the type directly [`5a2b738` (#356)](https://github.com/openwsn-berkeley/lakers/pull/356/commits/5a2b738c2dbed3e8a25d8b010edc0cc332ed992c)

shared/src/lib.rs
chrysn: Sorry, my comment was unclear: I didn't mean that it *needs* to be removed, I just observed that on platforms where it won't do anything sensible, the compiler *could* remove it.

But there's no harm in the current state either, because it's terribly unrealistic that we'll ever see an EDHOC message with a >4GiB EAD item :-)

PR #357, opened 2025-05-12
shared: Use EdhocBuffer as implementation for EdhocMessageBuffer

chrysn:
This removes a lot of duplicated code, and helps pave the way for using more suitably-sized buffers later.

I think it's really this easy, just creating a type alias … let's see if CI agrees with me.


examples/coap/src/bin/coapserver-coaphandler.rs
```diff
@@ -32,7 +32,7 @@ struct EdhocHandler {
 ///
 /// Places using this function may be simplified if From/Into is specified (possibly after
 /// enlarging the Error type)
-fn too_small(_e: MessageBufferError) -> Error {
+fn too_small(_e: EdhocBufferError) -> Error {
```
chrysn, line 35: This is the only public fall-out of the change -- and even that I'm changing here merely to avoid the deprecation warning.

In terms of semver compatibility, this is all a real breaking change, b/c two types that were previously distinct are now one, and that is visible for whoever impl'd any own trait on both MessageBufferError and EdhocBufferError (or the respective non-error types).

shared/src/lib.rs
```diff
@@ -560,120 +560,11 @@ pub enum CredentialTransfer {
...
-            Ok(EdhocMessageBuffer {
-                content: buffer,
-                len: self.len(),
-            })
-        } else {
-            Err(())
-        }
-    }
-}
+/// An [`EdhocBuffer`] used for messages (and, transitionally, other components).
```
chrysn, line 566: Should I elaborate on the "transitionally" here? (I think that'll become clear soon enough as we start not using a full EdhocMessageBuffer for everything, going forward from here).

chrysn: Yes; it's really up next, but doing that will just need a tad more testing compared to this (which is almost a binaries-should-be-identical PR).

PR #358, opened 2025-05-12
Update toolchain version

chrysn:
Dependencies (eg. icu_collections) start requiring newer stable Rust versions than our nightly had, breaking CI.

We don't have any MSRV guarantees in place right now, so "if it still builds it is fine" should apply to the nightly bump.


geonnave:
Thanks for this PR, agree we can bump it.

PR #359, opened 2025-05-14
Various small (possibly breaking) enhancements

chrysn:
On top of #357, this collects some little fixes I've gathered while exploring.

All should be straightforward to review as individual commits; some remarks on individual ones:

* Removing Default eases extensibility; moreover, all those contain public keys and/or MAC values, and the only reason those even implement Default is because they're type aliases and not newtypes -- there is no sensible "default" public key.

I'll switch this PR over to "ready to review" when there's nothing more to add in this chain -- but generally, if tests work and someone is eager to review and merge, be my guest.


chrysn:
I'm marking this as ready for review now.

If the hax parts are not as easy an update as I think, please let me know and I'll split the const parts out so that we can have this in rather sooner than later.

malishav:
`hax` check has been outdated for a while now, we need a separate effort in order to align that. I would be in favor of disabling the `hax` check until we get back to it. @geonnave thoughts?

chrysn:
I'd rather not throw hax overboard: Until we all get a feel again for what the next hax version can do, having it in here keeps us from falling into hard-to-verify patterns.

I can split off the const parts -- I'll eventually want to have them in, but that'll enable review on the other parts already, and the const can wait for when I'd add more test cases.

geonnave:
This all looks great, thanks @chrysn. 

As @malishav said, I would be in favor of merging the "constification" things over keeping hax.

If you choose to revert the const stuff and keep hax, I would recommend opening an issue where we can keep track of the "things that old hax is preventing us to do".

chrysn:
I've tried a bit of workarounds like the `const-ft` crate (which I'd have hoped to allow something like `#[cfg_attr(feature = "stable_consts", make_function_const)]`), but that doesn't work well because we'd wind up in macro hell (it's probably possible but very hard), so I'd take your suggestion and allow hax failures for now.

Seeing that the "merge" button is even active for me, this may not even need extra action: that check is failing (and that's kind of accurate), but that check is not set up as a required check anyway.

chrysn:
All that's left of this PR is the constifications (rest is already done in #364). I still think that this is something we should do, and will take you up on the "let's disregard hax for the time being", but I'll want to run a few things by the hax-as-it-was (trusting that this means it will also work with hax-as-it-will-be), and will keep this on hold a bit.

PR #360, opened 2025-05-22
Configurable EAD buffer sizes

chrysn:
This is a follow-up to https://github.com/openwsn-berkeley/lakers/pull/357#discussion_r2085078672 -- turns out there was really just one place where the EdhocMessageBuffer was (mis)used.

On the long run I'd prefer to get rid of this buffer altogether and instead work on owned message buffers (that might even be views onto an already buffered message), but that'll need larger API work, and this just changes types very locally.

### DIFF
--- a/lib/src/edhoc.rs
+++ b/lib/src/edhoc.rs
@@ -493,8 +493,8 @@ pub fn i_complete_without_message_4(state: &WaitM4) -> Result<Completed, EDHOCEr
     })
 }
 
-fn encode_ead_item(ead_1: &EADItem) -> Result<EdhocMessageBuffer, EDHOCError> {
-    let mut output = EdhocMessageBuffer::new();
+fn encode_ead_item(ead_1: &EADItem) -> Result<EADBuffer, EDHOCError> {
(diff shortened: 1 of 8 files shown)


lib/src/edhoc.rs
```diff
@@ -1737,8 +1737,8 @@ mod tests {
         let c_i_tv = C_I_TV;
 
         // the actual value will be zeroed since it doesn't matter in this test
-        let mut ead_value = EdhocMessageBuffer::new();
-        ead_value.len = MAX_MESSAGE_SIZE_LEN;
+        let mut ead_value = EdhocBuffer::new();
+        ead_value.len = MAX_EAD_LEN;
```
chrysn, line 1741: That I almost missed that is a strong indication that we should not have this pub in the first place; that's a panic bound to happen, especially when a user runs things on different size parametrizations than any hax panic-free proof runs on.

geonnave:
This is totally in line with what we want, thanks. (when I wrote the ead-authz code I was reluctantly re-using the EdhocMessageBuffer...)

PR #361, opened 2025-05-22
buffer: Close API

chrysn:
*Depends on: https://github.com/openwsn-berkeley/lakers/pull/365 (ignore the first few commits)*

The current shared buffer has all its members pub; that's not a good API design, especially as it means that an important invariant (len < CONST_LEN) is not upheld, as seen by an almost-bug in https://github.com/openwsn-berkeley/lakers/pull/360#discussion_r2102308069.

A large portion of the changes from this is that we go from

```
buffer.content[0] = 1;
buffer.content[1] = 2;
buffer.len = 2;
```

to

```
buffer.push(1).unwrap();
buffer.push(2).unwrap();
```

I'd prefer to have less unwrapping, but as a whole this is still easier to maintain (the maths got longer and longer in `buffer.content[thislenght + 2 + thatlength..thislength + 2 + thatlength + otherlength].copy_from_slice()`), and there was a possible panic implicit already (and we're running hax to tell us that it's fine). A better maintainable solution might be something like

```
let buffer = Buffer::<CONFIGURED_LENGTH>::new();
buffer.up_to::<20>.build()
    .push(1)
    .push(2)
    .extend_from_buffer(thing_with_a_maximum_length)
    .extend_from_array(hash);
buffer.extend_from_slice(slice)?;
```

where the up_to would be a const panic if CONFIGURED_LENGTH is insufficient, but that'd need some experimentation first.

<del>**Status**: So far, this is just preparatory commits (if we're to remove the direct pub access, we have to provide a few more methods), but I'd like to run them by CI step-by-step.</del>

### DIFF
--- a/lib/src/edhoc.rs
+++ b/lib/src/edhoc.rs
@@ -448,8 +448,7 @@ fn encode_ead_item(ead_1: &EADItem) -> Result<EADBuffer, EDHOCError> {
     };
 
     if let Some(label) = res {
-        output.content[0] = label;
-        output.len = 1;
+        output.push(label).unwrap();
 
         // encode value
         if let Some(ead_1_value) = &ead_1.value {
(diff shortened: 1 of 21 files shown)


lib/src/edhoc.rs
```diff
@@ -535,43 +534,34 @@ fn encode_message_1(
...
-    output.content[2 + raw_suites_len] = P256_ELEM_LEN as u8; // length of the byte string
-    output.content[3 + raw_suites_len..3 + raw_suites_len + P256_ELEM_LEN]
-        .copy_from_slice(&g_x[..]);
-    let c_i = c_i.as_cbor();
-    output.len = 3 + raw_suites_len + P256_ELEM_LEN + c_i.len();
-    output.content[3 + raw_suites_len + P256_ELEM_LEN..][..c_i.len()].copy_from_slice(c_i);
+    output.push(CBOR_BYTE_STRING).unwrap(); // CBOR byte string magic number
+    output.push(P256_ELEM_LEN as u8).unwrap(); // length of the byte string
+    output.extend_from_slice(&g_x[..]).unwrap();
+    output.extend_from_slice(c_i.as_cbor()).unwrap();
```
geonnave, line 564: very satisfying to get rid of `3 + raw_suites_len..3 + raw_suites_len + P256_ELEM_LEN` !!!

https://i.imgflip.com/xs4cs.jpg?a485520

geonnave:
Cool, I agree the .push API is already a pretty good improvement.

shared/src/lib.rs
```diff
@@ -499,7 +499,7 @@ pub struct ProcessingM2 {
     pub th_2: BytesHashLen,
     pub x: BytesP256ElemLen,
     pub g_y: BytesP256ElemLen,
-    pub plaintext_2: EdhocMessageBuffer,
+    pub plaintext_2: BufferPlaintext2,
```
geonnave, line 502: Checking my understanding: we go back to things like `BufferPlaintext2` so we can more easily slap a `EdhocBuffer<N>` on its type alias, whenever needed.

shared/src/buffer.rs
```diff
@@ -103,6 +115,33 @@ impl<const N: usize> EdhocBuffer<N> {
...
+    /// anticipating that the user will populate `self.content[result]`.
+    ///
+    /// ("Uninitialized" is in quotes because there are no guarentees on the content; from the
+    /// compiler's perspective, that area is initialized because this type doesn't play with
+    /// [`MaybeUninit`][core::mem::MaybeUninit], but don't rely on it).
+    ///
+    /// This is not a fully idiomatic Rust API: Preferably, this would return a `&mut [u8]` of the
+    /// requested length. However, as `.as_mut_slice()` or `.get_mut()` can not be checked by hax,
+    /// pushing and getting a range is the next best thing.
+    pub fn extend_reserve(
```
geonnave, line 128: Interesting, so this is to workaround around hax. Sounds good!

shared/src/lib.rs
chrysn: BufferPlaintext2 is already an EdhocBuffer type alias; the trouble is that some users of this expected it to be an EdhocMessageBuffer and some to be a BufferPlaintext2, and if those are not aliases (and we probably want to customize the N at some point, esp. if we want to prove panic freedom), we'll need to be precise even if the aliases happen to point to the same type when the consts align.

chrysn:
Thanks -- yes, ready to merge; ideally we merge 365 first though (it's a subset, and then it's clearer how things were reviewed)

PR #362, opened 2025-05-23
EXPERIMENTAL: Typestate buffer builder

chrysn:
As a result of #361, the pushes to the buffers in early message building (first few bytes, where we know things will fit) are very `.unwrap()`py. That is still OK there because it's an improvement over the explicit setting into content (that could also panic) and manual length setting (that could go wrong), but it's ugly.

Either kind of panic should be easy to optimize away, and for hax to prove, but there are still downsides:

* It's a visible panic that the developer has to convince themselves that it's OK.
* hax only runs on a particular size setting; can't cover all size choices in CI.

This approach is different in that it uses const generics (and occasionally typenum where const generics don't cut it, but for most parts that are user visible it's const generics) to perform a const assertion that the parts that are supposed to be infallible really are. (There's still a panic in there later, but if we trust the const generics we might even make that unsafe_unreachable, and at any rate, it's just *one* place to reason about).

This PR is currently in "initially run this by CI" stage, and not marked as a draft because I'd like feedback on the approach and whether we should follow this.


chrysn:
Rebased (and thus blocked by) #359. That already has the idiom of taking our `const {` blocks and hiding them from hax. All the other const'nesses in this PR are hopefully understood by hax.

chrysn:
This conflicts with #379 (hax checks) -- not just in the simple merge-conflict style, but because hax is unhappy with a lot of things in typenum, and can't do const generic expressions (which is an unstable feature anyway).

We might still offer the API under cfg(not(hax)), of offer an alternative (runtime panicking, but hax checkable) drop-in replacement.

PR #364, opened 2025-05-23
 Various small (but semver breaking) enhancements that do not break hax

chrysn:
This is the subset of #359 that does not affect hax.


geonnave:
All looks great!

Issue #363, opened 2025-05-23
Reduce use of buffers


chrysn:
Right now we're using various buffer type aliases about everywhere to lug around messages. This comes with a bunch of problems:

* We're copying around a lot of data, so much that we get huge stack sizes.
* There are type aliases all over the place but they're not used consistently, so they're decayed into essentially requiring them to be the same type.
* It's highly unidiomatic.

On the long run, I'd like to get to a point where we can do almost everything in place, provided hax lets us. But as a first step, let's get the use of buffers out from where we really don't need them, and clean up piece by piece.

## Failed attempts

I've tried to just make all the various type aliases of `EdhocMessageBuffer` be an `EdhocBuffer<{MAX_MESSAGE_SIZE + n}>`, but this failed horribly because we have started using them differently all over the place: For example, the crypto API expects a particular buffer that was the first user, but people put all kinds of buffer in, so we can't "just" go around and correct every type piece by piece because there's not a single correct type for every place.

chrysn:
A special case that's not easily seen when looking just at the Buffer occurrences are spots like edhco_kdf that already request a length and return an array (not a buffer) with the expectation that the caller pick their slice as requested. (The cases of "call with an array and a length of data in the array" are hopefully all gone, at latest with 3924aafb936842a21c738b0e382dcfe24d729b66 as currently in https://github.com/openwsn-berkeley/lakers/pull/361 that makes sha256_digest act on a slice). Searching for BytesMaxBuffer can help spot those. I'm not sure what to best replace them with; ideally I'd like to pass in a `&mut [u8]`; hax probably even tolerates that, but even better I'd like if that memory could be uninitialized (return write into slice … something has been brewing there in Rust, hasn't it?)

PR #365, opened 2025-05-23
Initial fixes for #363

chrysn:
This is WIP, with first commit only doing something similar to 4bd8715c9c8be6bfa871a4a0ba4290b47cadfad3 in #364


shared/src/lib.rs
```diff
@@ -499,7 +499,7 @@ pub struct ProcessingM2 {
     pub th_2: BytesHashLen,
     pub x: BytesP256ElemLen,
     pub g_y: BytesP256ElemLen,
-    pub plaintext_2: EdhocMessageBuffer,
+    pub plaintext_2: BufferPlaintext2,
```
geonnave, line 502: Is there a reason why some places are `plaintext: &[u8]` and others are `plaintext_2: BufferPlaintext2` ?

chrysn:
I think this is ready.

PR #369, opened 2025-06-03
EXPERIMENTAL: defmt: Implement Format unconditionally

chrysn:
This first iteration is mainly a test to see if we can do this unconditionally -- our CI builds Python binaries out of all those, so seeing it pass should be a good indication that indeed it's fine to implement defmt::Format even without having any linker script support.


chrysn:
The extra data defmt emits do not go well with the linker which the crates choose.

This might get better with https://github.com/knurling-rs/defmt/pull/878#issuecomment-2934936616 over time; let's retry when that one is through.

Issue #368, opened 2025-06-03
Remove most `repr(C)`


chrysn:
Many internal types are `repr(C)`. This is probably a relic from before the C bindings started copying everything into own structs that then are repr(C) -- but otherwise we don't do anything to warrant their use, so that repr can go away. (This is a breaking change, since repr(C) is an unsafety guarantee for others).

chrysn:
Un-holding it: the only thing this'd conflict with is the defmt experiments (https://github.com/openwsn-berkeley/lakers/pull/369), and they'll need a do-over anyway.

PR #370, opened 2025-06-04
Remove repr(C) from non-C-related code

chrysn:
Closes: https://github.com/openwsn-berkeley/lakers/issues/368

This is a breaking change, but we're there already. This might break things internally, but frankly I don't see how: We don't transmute, and all the C API bindings do have their copy_into_c style converters anyway.

Not making the types repr(C) allows the compiler to be smarter about them, and generally removes a promise we'd otherwise make and need to maintain.


chrysn:
A single item (`CredentialTransfer`) *was* used from the regular types also in the C FFI -- it's an enum with, for the foreseeable future, only two variants. Won't hurt much to just keep that repr(C).

PR #373, opened 2025-06-05
Upgrade manual buffers to use Buffer

chrysn:
After #361, there were a few places left where buffers were not used consistently and/or minimally:

* Some places still used manual buffer-ish constructs
* Some places used manual offset arithmetic rather than buffers to populate fixed-length outputs. Those were changed to use the more high-level idiom, which is easier to read and maintain while producing identical code thanks to const propagation and inlining.
* KDF / the EDHOC exporter returned maximally-sized (as per features, at least 256 bytes up to 1k) buffers, even though the caller passes in a length and needs to *do* something with that buffer anyway (i.e., all callers could have allocated, or usually already did allocate, a target buffer, and then copied from the maximal buffer into that target buffer). The APIs were changed to let the caller allocate.

There is still a use left of MAX_BUFFER_LEN, but that'll be another PR.

### DIFF
--- a/examples/coap/src/bin/coapclient.rs
+++ b/examples/coap/src/bin/coapclient.rs
@@ -77,8 +77,10 @@ fn client_handshake() -> Result<(), EDHOCError> {
     println!("EDHOC exchange successfully completed");
     println!("PRK_out: {:02x?}", prk_out);
 
-    let mut oscore_secret = initiator.edhoc_exporter(0u8, &[], 16); // label is 0
-    let mut oscore_salt = initiator.edhoc_exporter(1u8, &[], 8); // label is 1
+    let mut oscore_secret = [0; 16];
+    initiator.edhoc_exporter(0u8, &[], &mut oscore_secret); // label is 0
+    let mut oscore_salt = [0; 8];
+    initiator.edhoc_exporter(1u8, &[], &mut oscore_salt); // label is 1
 
     println!("OSCORE secret: {:02x?}", oscore_secret);
     println!("OSCORE salt: {:02x?}", oscore_salt);
@@ -92,8 +94,8 @@ fn client_handshake() -> Result<(), EDHOCError> {
     println!("PRK_out after key update: {:02x?}?", prk_out_new);
 
     // compute OSCORE secret and salt after key update
-    oscore_secret = initiator.edhoc_exporter(0u8, &[], 16); // label is 0
-    oscore_salt = initiator.edhoc_exporter(1u8, &[], 8); // label is 1
+    initiator.edhoc_exporter(0u8, &[], &mut oscore_secret); // label is 0
+    initiator.edhoc_exporter(1u8, &[], &mut oscore_salt); // label is 1
 
     println!("OSCORE secret after key update: {:02x?}", oscore_secret);
     println!("OSCORE salt after key update: {:02x?}", oscore_salt);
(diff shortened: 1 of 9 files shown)


crypto/lakers-crypto-psa/src/lib.rs
```diff
@@ -55,19 +55,18 @@ impl CryptoTrait for Crypto {
...
 
         for i in 2..=n {
             message[..SHA256_DIGEST_LEN].copy_from_slice(&t_i);
             message[SHA256_DIGEST_LEN..SHA256_DIGEST_LEN + info.len()].copy_from_slice(&info);
             message[SHA256_DIGEST_LEN + info.len()] = i as u8;
             t_i = self.hmac_sha256(&message[..SHA256_DIGEST_LEN + info.len() + 1], prk);
-            output[(i - 1) * SHA256_DIGEST_LEN..i * SHA256_DIGEST_LEN].copy_from_slice(&t_i);
+            let start = (i - 1) * SHA256_DIGEST_LEN;
+            let t_i_len = core::cmp::min(result[start..].len(), SHA256_DIGEST_LEN);
+            result[start..start + t_i_len].copy_from_slice(&t_i[..t_i_len]);
```
malishav, line 68: Could you remind me if we have explicit test vectors that test HKDF-expand and -extract? I suppose CI not complaining means that the changes are OK.

crypto/src/lib.rs
```diff
@@ -77,7 +77,9 @@ mod tests {
         let mut info = [0; MAX_INFO_LEN];
         info[..info_slice.len()].copy_from_slice(info_slice);
 
-        let okm = crypto.hkdf_expand(&prk, &info[..info_slice.len()], output_length);
-        assert_eq!(okm[..output_length], expected_okm_slice[..]);
+        let mut output = Vec::new();
```
malishav, line 80: Is this the only use of `Vec` in the codebase? We have so far avoided them.

malishav (commented):
This PR looks great, so much simplification! I have a question on the use of `Vec` though, which we have avoided so far.

crypto/lakers-crypto-psa/src/lib.rs
chrysn: I don't know if we have any; they are generally covered by the fuller EDHOC tests.

For this case specifically (the extraction output being longer than a single block), I'll do a run where the output gets deliberately broken to see if that case is covered too. (Usually we'd try to avoid that case, to the point where I've seen HKDF implementations that don't even support the more-than-one-block form).

chrysn: Local tests broke with

```diff
diff --git a/crypto/lakers-crypto-rustcrypto/src/lib.rs b/crypto/lakers-crypto-rustcrypto/src/lib.rs
index 93a7709..cabc828 100644
--- a/crypto/lakers-crypto-rustcrypto/src/lib.rs
+++ b/crypto/lakers-crypto-rustcrypto/src/lib.rs
@@ -54,2 +54,5 @@ impl<Rng: rand_core::RngCore + rand_core::CryptoRng> CryptoTrait for Crypto<Rng>
             .expect("Static lengths match the algorithm");
+        if result.len() > 32 {
+            result[result.len() - 1] ^= 0x55;
+        }
     }
```

in the test `tests::test_hkdf_expand::case_2` -- so yes, we do have a test case covering those.

PR #379, opened 2025-06-10
buffer: Towards hax typechecking

chrysn:
These changes do a few things in buffer around hax:

* [x] make hax lax typechecking pass
* [x] make hax strict typechecking pass
* [x] explain why the length in the buffer is not annotated as refined
* [x] add the hax-lib module so that we can do some annotations
* [x] remove code that typechecking showed to be wrong anyway
* [x] makes .len not pub (because while that doesn't stop typechecking for some reasons, it'd defeat its purpose)

We could go a little less of the way and not depend on hax-lib if we're happy with the workarounds in the "Bend unreachable cases for hax to understand" commit.

They don't yet [edit: i.e, this was open at the time the PR was started]

* [x] enforce that hax typechecking stays passing
* [x] let hax pass as a whole (because since the Index impl doesn't typecheck, it's disabled for hax, and now unavailable for its users)
* [x] pass normal tests (oups, where did *that* go wrong?)

Is this a direction we want to go?

### DIFF
--- a/shared/src/buffer.rs
+++ b/shared/src/buffer.rs
@@ -86,11 +86,11 @@ impl<const N: usize> EdhocBuffer<N> {
         }
     }
 
-    pub fn len(&self) -> usize {
+    pub const fn len(&self) -> usize {
         self.len
     }
 
-    pub fn capacity(&self) -> usize {
+    pub const fn capacity(&self) -> usize {
         N
     }
 


shared/src/buffer.rs
```diff
@@ -4,6 +4,42 @@ use core::ops::Index;
...
+        //
+        // Forcing it to convince itself that the length stays the same is not just relevant for
+        // the loop (where the assignment would fail otherwise), but also for later where the
+        // function is used: Only with this, the "returned" version of long will even "fit" in the
+        // slot where it was taken from under hax' move-calling pattern.
+        hax_lib::loop_invariant!(cursor <= short.len() && long.len() == _original_length);
+        cursor = cursor - 1;
+        // This assert triggers after assigning -- doesn't the item assign operation
+        // provide that the length stays the same? That it was true was a precondition!
+        //assert!(short.len() <= long.len());
```
chrysn, line 38:
```suggestion
```

(Leftover from finding out why things go wrong)

malishav (approved):
This is absolutely fantastic, thank you so much for providing this PR!!!

PR #380, opened 2025-06-24
EAD list tuning

chrysn:
This PR takes up a buch of suggestions I had for #356.

I'll go through them in detail and ping reviewers once done; now I'm just doing occasional CI runs on it.

### DIFF
--- a/ead/lakers-ead-authz/src/device.rs
+++ b/ead/lakers-ead-authz/src/device.rs
@@ -71,14 +71,14 @@ impl ZeroTouchDeviceWaitEAD2 {
     pub fn process_ead_2<Crypto: CryptoTrait>(
         &self,
         crypto: &mut Crypto,
-        ead_2: EADItem,
+        ead_2: &EADItem,
         cred_v: &[u8],
     ) -> Result<ZeroTouchDeviceDone, ZeroTouchError> {
         trace!("Enter process_ead_2");
         if ead_2.label != EAD_AUTHZ_LABEL {
             return Err(ZeroTouchError::InvalidEADLabel);
         }
-        let Some(ead_2_value_buffer) = ead_2.value else {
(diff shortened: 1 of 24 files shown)


chrysn:
This is now in a state in which it could be reviewed. As it builds on work by @WilliamTakeshi, could you have a coarse look at whether I changed something that you had good reasons to do this way?

The commits should be self-explanatory; notes where I think they add:
* Types were renamed because in Ead and EADItem it was not clear what is what.
* Internals were made non-pub; generally I think this library has too many pub details, and worse here, things could get inconsistent between how long the list is and where there is actually Some item.
  After hiding the implementation details, I could remove the `.len`, as that enables one of the two patterns in which I think this will be realistically used:
  - By iteration: For this, the idiomatic `.iter()` API is now provided.
  - By removing items: An application can consist of subsystems, all of which are fed the EAD list and they get a chance to update the application's decisions taken. The pattern of `.pop_by_label()` allows them to independently pluck their EAD items from the list, and more importantly, the application can then (even in presence of rather independent subsystems) reliably come to a conclusion on whether all critical items ahve been processed or not.
  - For Python, having a list saves us the hassle of implementing all the things that a list-ish thing *should* have in Python -- instead, we just *return* a list and don't expose the EadItems type. The downside is that the `.pop_by_label()` method is not available, but Python users can use that pattern in a similar way by manipulating lists.
* Minor API changes on how authz takes in EAD items simplified using it with the iterator pattern -- and at any rate, why pass a big thing by value if it's not modified or otherwise taken.

There's a final fix I'll have to apply to make hax happy, but that shouldn't stand in the way of reviewing this as a whole.

chrysn:
Placing this on the priority list as now both #372 and further work in #381 are blocked on this.

WilliamTakeshi (approved):
Looks good to me — thanks for taking the time to improve this part!

PR #381, opened 2025-06-25
ead: Don't make caller allocate full buffer

chrysn:
Depends on: #380

It's great that we can now use multiple EAD items, but our style so far of always passing things by value starts to bite now: Our stack usage is already larger than it should be, and now the creator of a message has to allocate MAX_EAD_LEN * MAX_EAD_ITEMS on its stack just so that it can populate a buffer where those would fit in anyway. Ideally, it'd even zero-copy populate the EADs into a buffer, but let's start "small" and allow the caller (eg. of prepare_message2) to allocate what it needs and then pass that in.

The commit currently in here is a test balloon only running this for prepare_message_2; eventually, this will need to be consistent across all prepare_message_n. Further steps from there:

* [ ] See if hax likes it.
* [ ] Thread it through more widely caller-side (currently, in many cases there are EadItems created that are only later mapped into an iterator of EadSlice; ergonomics would be better when creating an EadSlice right away)
* [ ] Thread it through for all messages (1, 3 and 4)
* [ ] See if we can do the same on the receiver side too, handing out slices of the received (and buffered) message.


chrysn:
Preliminary checks sound good: Both in entering EAD2 and in reading EAD3, these avoid ever having a full max-length*max-items buffer -- and checks still pass.

There are open concerns still around:
* [ ] Ergonomics of users splitting out handling of different EAD items and eventually checking that all critical ones are gone. This is a regression compared to the not-even-merged-yet https://github.com/openwsn-berkeley/lakers/pull/380, and options are to build "iterator but one option is removed" types from type state or from runtime state. Or we just leave it as it was before #380, and the user needs to convince themself that no critical options remain. (BTW, this is something CoAP libraries often have too, with different approaches in my Rust and the RIOT CoAP implementation).
* [ ] In some places (at least Python API), the current code rebuilds a stack-allocated full EAD item list -- but that's just an exercise in making the change vs. taking the easiest-to-change route.
* [ ] On the long run, storing the EAD items in a slice (as is effectively done between parsing and verifying message 3) for later iteration rather than parsing it upfront raises questions of panics: The first code to process the message should (but doesn't do yet) verify that the EAD (CDDL `*(int, ?bytes)`) format is followed, and when later iterating, any format violation is a panic but it's really unreachable because the type should not be constructable in an unchecked way. Letting hax follow this might require formalizing CBOR validity in hax, which is an interesting exercise anyhow. (I don't expect to do that for here so far; right now we don't even typecheck that module).

chrysn:
@geonnave, could you give me a high-level check of whether you see any reason not to go through with this?

Issue #329, opened 2025-01-16
EAD labels > 23 are misencoded, discarding criticality


chrysn:
I think that the only label/criticality combinations that are currently encoded right are ([1..23], false).

While the decoder properly distinguishes between u8 and i8 values, and can at least decode the simple ones right. However, encoding 25 results in a parser error later.

[edit:] Adding some context: Imlpementing draft-ietf-ace-edhoc-oscore-profile-06, I have to use some value for the "ACE-OAuth Access Token" item. As no early allocation is happening yet, I'll have to make a guess, and I'd have guessed 25. I guess I'll update my guess to 20.

CC'ing @malishav who is DE for that registry and may want to earmark something.

Issue #382, opened 2025-07-02
Voucher is double-encoded in bytes


chrysn:
The voucher response as is part of our test vectors contains the CBOR encoded byte string of the CBOR encoded byte string of the 8-byte MAC; in the test vectors, that shows up as `4948`.

The reference document so far is draft-ietf-lake-authz-01; note that a few things have changed in the meantime. Still, the voucher is supposed to be `Voucher = COSE_Encrypt0.ciphertext`, and that is in our case just a MAC, which should be 8 bytes -- instead, voucher is transported as the byte string around those 8 bytes (making it 9 bytes).

While refactoring, I won't change the test vectors, and will leave things functionally as they are, but as EADItem starts to store the slice rather than the encoded slice (for doing the latter creates a hard-to-enforce invariant on a structure that has all its fields pub), things that have looked smooth start looking weird due to the extra layer, and this issue helps keep track of them by providing an identifier associated with those places in the code.

PR #383, opened 2025-07-02
EAD: Process full range of labels (and more refactoring)

chrysn:
This started out as just extending the range to enable GREASE, but escalated into refactoring that keeps applications from inserting invalid CBOR into messages, circling back to enabling the range not just for decoding but also for encoding.

Some notes:
* Generally (and this is not being changed here), the values of EADItems are stored in CBOR encoded form, i.e. as on the wire -- which is a previously undocumented invariant on the stored value field: It must contain zero or one encoded byte strings, otherwise it won't round-trip through serialization.
  - Once we typecheck shared, this invariant will need to be expressed -- but maybe we'll use a completely different internal representation by then.
  - Switching to only storing the bytes and not the encoded length is something I'd consider, but that'll need trying out, and having the struct fields private (see below) is a prerequisite for doing that.
* There were some oddities around what it means for an EADItem to have a value or not, as it used to be stored in an `Option<Buffer>`. In particular, items without a value used to be skipped from some operation (even their labels were ignored), but then again, in no place where they were actually constructed was the option None.
  - For consistency, the value is now only stored in a buffer, which is empty when there is no value. The invariant remains that that buffer contains zero or one encoded byte string.
  - Things got very weird where the voucher was sent from W to V in a way that only by W being very trustworthy would not mess up the encoding -- at least so it seems from https://github.com/openwsn-berkeley/lakers/issues/382. API-wise, this necessitated a few changes that still work with the test vectors, but given that that part of ELA changed significantly in the meantime, I didn't spend too much time making things pretty.
* Internals of EADItem are made private. This enables enforcing the invariant, and enables future experiments as to what the best internal representation of all this is.
* EAD encoding is methods of EAD, and makes much more sense in shared where the type is defined -- code has been moved, and that allowed deduplication. (Previously, the loop to encode everything that is in an `EadItems` was spelled out 5 times).

Closes: https://github.com/openwsn-berkeley/lakers/issues/329


chrysn:
In terms of PR interaction, this follows the theme of #380 of cleaning up EADs (380 enhanced the handling of EAD lists, this is more on the EADs themselves).

This will probably conflict terribly with #381, which now remains as a demo and request for "do we want to do this properly", but I'll likely have to do that one over again once this one is discussed and merged.

chrysn:
Builds failed because the last step of "let's hide the fact that we store things in encoded form completely" broke C. Took the opportunity to make the C and Python APIs use the consistent view new view of "the user never needs to do CBOR encoding or decode CBOR" (unless, of course, they / their application chooses to use CBOR inside that byte string).

malishav (approved):
I went over this PR which looks great and IMO ready to merge! it would be great if @geonnave could go over it as well, as it touches the codebase of lake-authz quite a bit.

lakers-c/src/lib.rs
```diff
@@ -26,26 +26,32 @@ static HEAP: Heap = Heap::empty();
 pub struct EADItemC {
     pub label: u16,
     pub is_critical: bool,
+    /// The value is only emitted if this is true (otherwise it is an EAD item is just a label)
```
geonnave, line 29: do you mean: otherwise it is *NOT* an EAD item (?)

shared/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub const AES_CCM_TAG_LEN: usize = 8;
 pub const MAC_LENGTH: usize = 8; // used for EAD Zeroconf
```
geonnave, line 62: NOTE: stopped review here. To continue from this file

lakers-c/src/lib.rs
chrysn:
```suggestion
    /// The value is only emitted if this is true (otherwise it is an EAD item that has just a label)
```

It's still an EAD item, but the wording was wrong.

ead/lakers-ead-authz/src/server.rs
```diff
@@ -166,8 +166,11 @@ fn encode_voucher_response(
     output.extend_from_slice(message_1.as_slice()).unwrap();
 
     output
-        .push(CBOR_MAJOR_BYTE_STRING + ENCODED_VOUCHER_LEN as u8)
+        .push(CBOR_MAJOR_BYTE_STRING + 1 + VOUCHER_LEN as u8)
         .unwrap();
+    // The voucher is double-wrapped in bytes; see
+    // <https://github.com/openwsn-berkeley/lakers/issues/382>
```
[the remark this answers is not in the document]
chrysn:
```suggestion
    // <https://github.com/lake-rs/lakers/issues/382>
```

There are redirects in place, so it's not wrong, but as I'm doing a documentation fix anyway, I can change that too.

geonnave:
Finished reviewing and all looks good. Feel free to merge!

chrysn:
Thanks.  I'm merging the PyO3 update changes (which have since been merged) back in, to get a clearly trackable history of which parts change where due to the merge. If CI passes, I'll merge that into main.

PR #385, opened 2025-07-15
Update pyo3

chrysn:
Our Python package is not compatible with the latest PyPy versions (and probably also not with upcoming Python versions) due to its old pyo3. ("error: the configured PyPy interpreter version (3.11) is newer than PyO3's maximum supported version (3.10)").

An update to pyo3 should fix this; that happens in two steps to get the benefits of deprecation warnings during development (and thus also in any bisection situation).


geonnave:
Went through and it looks good to me!

PR #386, opened 2025-07-19
lakers-c: handle case when EAD param is null

geonnave:
And also apply some formatting comments on the headers generated by cbindgen.


chrysn:
Can we check this after https://github.com/lake-rs/lakers/pull/383 ? The EAD API is still just settling…

chrysn (approved):
But really this looks orthogonal to the changes there, so we can merge it right away too. Looks straightforward enough. There is no documentation for that function, so it's a valid assumption that a C user might pass in a null pointer. (We should probably have docs, but then, I wouldn't spend too much time on the C side at this level).

geonnave:
Sure! I sent this PR as I got bitten when using the C API. I also realize now I am owing a review, will do it during the IETF Hackathon.

geonnave:
Since it's orthogonal -- merging!

PR #415, opened 2026-03-12
Refactoring adding psk method

ElsaLopez133:
Adding PSK method after refactoring.

- `edhoc.rs` is kept as the main EDHOC protocol file and depending on the method we branch to different functions that are stored in `edhoc/statstat.rs` and `edhoc/psk.rs`. Therefore,  `edhoc/statstat.rs` and `edhoc/psk.rs` contain only the parts that differ between the Static-Static and PSK methods.

- These method-specific files return result structs back to `edhoc.rs`, so the central flow can continue without duplicating the shared protocol logic.

- We store method-specific fields in `enums` such as `WaitM3MethodSpecifics`, `ProcessingM2MethodSpecifics`,` ProcessedM2MethodSpecifics`, and `ProcessingM3MethodSpecifics`. These enums allow the common state structs to carry only the extra data required by the selected method. The use of  `enums`  was chosen to avoid having `Option` values.


lib/src/edhoc.rs
```diff
@@ -98,7 +104,27 @@ pub fn r_parse_message_3(
...
+}
+
+pub fn r_parse_message_3_with_cred_resolver<F>(
+    state: &mut WaitM3,
+    crypto: &mut impl CryptoTrait,
+    message_3: &BufferMessage3,
+    resolve_cred_i: F,
+) -> Result<(ProcessingM3, IdCred, EadItems), EDHOCError>
+where
+    F: Fn(&IdCred) -> Result<Credential, EDHOCError>,
```
chrysn: So far we've managed to communicate with the application without the need for callbacks, because they'd lead us to function-coloring hell -- what if that lookup accessed the disk or EEPROM or network and thus needed to be async?

(Also, the callback should be FnOnce, but we shouldn't let it come to that even.)

crypto/lakers-crypto-psa/src/lib.rs
```diff
@@ -88,6 +88,14 @@ impl CryptoTrait for Crypto {
         output
     }
 
+    fn hkdf_extract_psk(&mut self, salt: &BytesHashLen, ikm: &BytesElemLenPSK) -> BytesHashLen {
+        // TODO
+        // TODO generalize if salt is not provided
+        let output = self.hmac_sha256(&mut ikm.clone()[..], salt);
```
WilliamTakeshi: I think this can be simplified to `self.hmac_sha256(ikm, salt);`?

shared/src/crypto.rs
```diff
@@ -61,6 +61,7 @@ pub trait Crypto: core::fmt::Debug {
     fn sha256_start<'a>(&'a mut self) -> Self::HashInProcess<'a>;
     fn hkdf_expand(&mut self, prk: &BytesHashLen, info: &[u8], result: &mut [u8]);
     fn hkdf_extract(&mut self, salt: &BytesHashLen, ikm: &BytesP256ElemLen) -> BytesHashLen;
+    fn hkdf_extract_psk(&mut self, salt: &BytesHashLen, ikm: &BytesElemLenPSK) -> BytesHashLen;
```
WilliamTakeshi, line 64: if you want, you can make this generic.

but I think you can keep like this and we fix on embedded-cal

examples/coap/src/bin/coapclient.rs
```diff
@@ -54,8 +54,10 @@ fn client_handshake() -> Result<(), EDHOCError> {
     println!("message_2 len = {}", response.message.payload.len());
 
     let message_2 = EdhocBuffer::new_from_slice(&response.message.payload[..]).unwrap();
-    let (mut initiator, c_r, details, ead_2) = initiator.parse_message_2(&message_2)?;
-    let ParsedMessage2Details::StatStat { id_cred_r } = details;
+    let (mut initiator, c_r, details) = initiator.parse_message_2(&message_2)?;
+    let ParsedMessage2Details::StatStat { id_cred_r, ead_2 } = details else {
+        return Err(EDHOCError::UnsupportedMethod);
+    };
```
WilliamTakeshi: Just thinking out loud.

Isn't weird that we give the responsability to checking the details to the user? initiator.parse_message_2 should just handle everything and the user would just process/parse and receive the `Err` back.

Not that we are adding more METHODS, maybe we can discuss later how we want the user to use the API, as of today I think he has to do many checks that should be made internally

lakers-c/src/initiator.rs
```diff
@@ -141,7 +142,8 @@ pub unsafe extern "C" fn initiator_verify_message_2(
     let crypto = &mut default_crypto();
 
     let state = core::ptr::read(&(*initiator_c).processing_m2).to_rust();
-
+    
+    // FIXME
```
WilliamTakeshi: What needs to be fixed?

lakers-c/src/initiator.rs
```diff
@@ -106,11 +105,15 @@ pub unsafe extern "C" fn initiator_parse_message_2(
...
+        Ok((state, c_r, details, ead_2)) => {
             ProcessingM2C::copy_into_c(state, &mut (*initiator_c).processing_m2);
             let c_r = c_r.as_slice();
             assert_eq!(c_r.len(), 1, "C API only supports short C_R");
             *c_r_out = c_r[0];
+            let id_cred_r = match details {
+                ParsedMessage2Details::StatStat { id_cred_r } => id_cred_r,
+                ParsedMessage2Details::Psk {} => IdCred::new(),
+            };
             *id_cred_r_out = id_cred_r;
```
WilliamTakeshi: on the PSK case, why write an empty `id_cred_r` in `id_cred_r_out`. I thought we would just do nothing.

lib/src/edhoc/psk.rs
```diff
@@ -0,0 +1,232 @@
...
+    F: Fn(&IdCred) -> Result<Credential, EDHOCError>,
+{
+    let res = parse_message_3(message_3)?;
+    let plaintext_3a = encrypt_decrypt_ciphertext_3a(crypto, &state.prk_3e2m, &state.th_3, &res);
+    let id_cred_psk = IdCred::from_encoded_value(&[plaintext_3a.as_slice()[0]])?;
+
+    let cred_i = resolve_cred_i(&id_cred_psk)?;
+
+    let mut ciphertext_3b = BufferCiphertext3::new();
+    let _ = ciphertext_3b.fill_with_slice(&plaintext_3a.as_slice()[1..]);
```
WilliamTakeshi: here we are silently discarding errors. is this what we want? Check if we want to return the error on this case

lib/src/edhoc/psk.rs
```diff
@@ -0,0 +1,232 @@
...
+    let mut ct_3a: BufferCiphertext3 = BufferCiphertext3::new();
+    ct_3a.extend_from_slice(pt_3a).unwrap();
+    ct_3a.extend_from_slice(ciphertext_3b.as_slice()).unwrap();
+    let ciphertext_3a = encrypt_decrypt_ciphertext_3a(crypto, &state.prk_3e2m, &state.th_3, &ct_3a);
+    // CBOR encoding of ct_3a
+    let encoded_ciphertext_3a = encode_ciphertext_3a(ciphertext_3a)?;
+    //compute message_3
+    message_3
+        .extend_from_slice(encoded_ciphertext_3a.as_slice())
+        .unwrap();
```
WilliamTakeshi: we have 3 unwrap() on this function, but it returns a Result<>, can they be errors?

lib/src/edhoc/psk.rs
```diff
@@ -0,0 +1,232 @@
...
+    message_3: &BufferMessage3,
+    cred_r: &Credential,
+    resolve_cred_i: F,
+) -> Result<ParsedMessage3, EDHOCError>
+where
+    F: Fn(&IdCred) -> Result<Credential, EDHOCError>,
+{
+    let res = parse_message_3(message_3)?;
+    let plaintext_3a = encrypt_decrypt_ciphertext_3a(crypto, &state.prk_3e2m, &state.th_3, &res);
+    let id_cred_psk = IdCred::from_encoded_value(&[plaintext_3a.as_slice()[0]])?;
```
WilliamTakeshi: can you check if when you create a empty message3, if this will panic. Also, are we sure that we only want to send the first byte? because reading the from_encoded_value, it may accept len() == 2. Because I don't know PSK well, I am not sure about this comment

lib/src/edhoc.rs
```diff
@@ -596,28 +649,57 @@ fn encode_plaintext_4(ead_4: &EadItems) -> Result<BufferPlaintext4, EDHOCError>
...
-    enc_structure.push(SHA256_DIGEST_LEN as u8).unwrap();
-    enc_structure.extend_from_slice(th_3).unwrap();
+    enc_structure.extend_from_slice(encrypt0).unwrap();
+
+    // protected field: zero-length byte string
+    enc_structure.push(CBOR_MAJOR_BYTE_STRING | 0).unwrap();
+
+    // external_aad field
+    enc_structure.push(CBOR_BYTE_STRING).unwrap();
+    enc_structure.push(external_aad.len() as u8).unwrap();
```
WilliamTakeshi: is it possible that external_aad.len()>255? if thats possible, this fails. because in rust `as` wraps the value, so lets say its 256, it wiill write 0

lib/README.md
```diff
@@ -42,7 +42,10 @@ let initiator = EdhocInitiator::new(default_crypto());
 
 let (initiator, message_1) = initiator.prepare_message_1(None, &EADItem::new_empty_array())?; // c_i and ead_1 are set to None
 
-let (initiator, _c_r, id_cred_r, _ead_2) = initiator.parse_message_2(&message_2)?;
+let (initiator, _c_r, details) = initiator.parse_message_2(&message_2)?;
```
WilliamTakeshi: parse_message_2 returns 4 items,

lakers-c/src/initiator.rs
WilliamTakeshi: ok, maybe I got it. its hardcoded to statstat

lib/src/lib.rs
```diff
@@ -272,7 +326,7 @@ impl<'a, Crypto: CryptoTrait> EdhocInitiator<Crypto> {
     }
 
     pub fn set_identity(&mut self, i: BytesP256ElemLen, cred_i: Credential) {
-        self.i = Some(i);
+        self.i = Some(InitiatorIdentity::StatStat { i });
```
WilliamTakeshi: do you need a set_identity for PSK? or its the default?

lib/src/edhoc/psk.rs
```diff
@@ -0,0 +1,232 @@
...
+    BufferCiphertext3, BufferMessage3, BufferPlaintext2, BytesHashLen, ConnId, Credential,
+    CredentialKey, CredentialTransfer, DecodedMessage2, EDHOCError, EadItems, IdCred,
+    ParsedMessage2Details, ParsedMessage3, PreparedMessage2, PreparedMessage3, ProcessedM2,
+    ProcessedM2MethodSpecifics, ProcessingM2, ProcessingM2MethodSpecifics, ProcessingM3,
+    ProcessingM3MethodSpecifics, Th4Input, VerifiedMessage2, VerifiedMessage3, WaitM3,
+    WaitM3MethodSpecifics,
+};
+use lakers_shared::Crypto as CryptoTrait;
+
+pub fn r_prepare_message_2_psk(
```
WilliamTakeshi: maybe they should be pub(crate)? please check the whole psk.rs if they are being used outside and limit the API surface as much as you can

lib/src/edhoc.rs
```diff
@@ -952,6 +1077,14 @@ fn compute_prk_4e3m(
     crypto.hkdf_extract(salt_4e3m, &g_iy)
 }
 
+fn compute_prk_4e3m_psk(
+    crypto: &mut impl CryptoTrait,
+    salt_4e3m: &BytesHashLen,
+    cred: &BytesKeyAES128, //TODO: what is the psk type? len?
```
WilliamTakeshi: Is this type right?

I thought cred was a &BytesElemLenPSK. at least is how its defined on  `crypto.hkdf_extract_psk`. They are different types (but because on a lower level they are the same, the compiler doesn't complain)

WilliamTakeshi (commented):
Besides the comments I put here,

1. can you also check if nRF52 code still runs??? I don't have one with me right now :((
2. write an full e2e test like test_handshake(). It will show if everything works and give an example for future. Also you can use to test, like the one that I said when message3 is empty you can force the responder to crash

shared/src/crypto.rs
ElsaLopez133: Yes, i actually thought of doing `fn hkdf_extract(&mut self, salt: &BytesHashLen, ikm: &[u8]) -> BytesHashLen;` but then that means that the function accepts any length for ikm? Not sure if this is okay, so that is why I defined another constant to capture the length of the PSK

examples/coap/src/bin/coapclient.rs
ElsaLopez133: I think that makes sense, specially because that information is already part of state.method_specifics, so no need to expose it in the function as well. Will update!

lakers-c/src/initiator.rs
ElsaLopez133: So I addressed that by adding a flag `has_id_cred_r_out`. The flag now tells  whether ID_CRED_R is present. I still initialize id_cred_r_out to a default empty value in the PSK case just to avoid leaving output memory undefined on the C side, but maybe this is not strictly needed?

lib/src/edhoc/psk.rs
ElsaLopez133: True. Updated it to
```
    ciphertext_3b
        .fill_with_slice(&plaintext_3a.as_slice()[1..])
        .map_err(|_| EDHOCError::ParsingError)?;
```
Still unsure about what is the best error in this case...

lib/src/edhoc/psk.rs
ElsaLopez133: Nice comment. This is addressed now. The empty message_3 case no longer panics: parse_message_3() checks for empty input and returns EDHOCError::ParsingError instead (didnt think of this).

I also removed the assumption that ID_CRED_PSK is always one byte. Instead of taking only plaintext_3a[0], the PSK path now uses CBORDecoder::any_as_encoded() to parse the first full CBOR item from plaintext_3a, and then feeds that into IdCred::from_encoded_value(). That means longer valid encoded ID_CRED_PSK forms are handled correctly now. Thanks!

lib/src/edhoc.rs
ElsaLopez133: True. That assumption held when external_aad was just TH_3, but not anymore with PSK because the additional fields can push it past 255 bytes. I fixed it by adding an encode_bstr_header helper that encodes the CBOR bstr length correctly for the actual size

lib/src/edhoc/psk.rs
ElsaLopez133: Yes, they should. I forgot to implement the changes made on statstat.rs

lib/src/lib.rs
ElsaLopez133: Yes, we still need set_identity for PSK. In the PSK case there is no P-256 i value to provide, but set_identity is still how the initiator records that it is using InitiatorIdentity::Psk and stores cred_i, which is required later when preparing message 3.

lib/src/edhoc.rs
ElsaLopez133: Fixed. This was defined when I still did not have the BytesElemLenPSK, but yes, it should be that.

lakers-c/src/initiator.rs
WilliamTakeshi: Good question, I don't know what is the right thing to do on C, but I guess better than adding a flag `has_id_cred_r_out` it should be something like `method` and be a `3` for Statstat and `4` for Psk. 

And yes, reading again I think your code is correct, they sohuld still have the empty value initialized in the PSK case, we just ignore it.

lib/src/edhoc/psk.rs
WilliamTakeshi: Also it should change the [1..] a couple lines below, I think

lib/src/edhoc.rs
```diff
@@ -892,22 +1015,35 @@ fn encode_plaintext_2(
...
     ead_2.encode(&mut plaintext_2)?;
 
     Ok(plaintext_2)
 }
 
+fn encode_ciphertext_3a(ciphertext: EdhocMessageBuffer) -> Result<BufferCiphertext3, EDHOCError> {
+    let mut ciphertext_3a: BufferCiphertext3 = BufferCiphertext3::new();
+    // plaintext_3a: P = ( ID_CRED_PSK / bstr / int )
+    ciphertext_3a
+        .push(CBOR_MAJOR_BYTE_STRING | ciphertext.len() as u8)
```
WilliamTakeshi: This is wrong for `ciphertext.len() > 23`. Is it possible that the `ciphertext.len()` is greater than 23?

I think you can use `encode_bstr_header`

lib/src/edhoc.rs
```diff
@@ -892,22 +1015,35 @@ fn encode_plaintext_2(
...
+        .or(Err(EDHOCError::EncodingError))?;
+    let _ = ciphertext_3a.extend_from_slice(ciphertext.as_slice());
```
WilliamTakeshi: Here we are ignoring in case of errors. its better to return an `Err`

lakers-python/src/responder.rs
```diff
@@ -39,13 +39,21 @@ pub(crate) enum PyEdhocResponderSummary {
 #[pymethods]
 impl PyEdhocResponder {
     #[new]
-    fn new(r: Vec<u8>, py: Python<'_>, cred_r: super::AutoCredential) -> PyResult<Self> {
+    #[pyo3(signature = (cred_r, r=None))]
+    fn new(py: Python<'_>, cred_r: super::AutoCredential, r: Option<Vec<u8>>) -> PyResult<Self> {
```
WilliamTakeshi, line 43: Here we are breaking the API because we are changing the order, we have to document this somewhere, because it will break people updating the Python API. I don't know if this is fine.

WilliamTakeshi, line 43: Also another issue is, `Option<Vec<u8>>` "represents" 3 states:

1. Not provided - `None`
2. Provided wo data - `Some([])`
3. Provided with data - `Some([1,2])`

Its not clear the difference between the user using the API with 1. or 2.


But I am fine keeping like this so we can merge PSK and thinking the APIs later :))

lakers-python/src/responder.rs
```diff
@@ -321,3 +358,14 @@ impl PyEdhocResponder {
...
     }
 }
+
+fn parse_responder_identity(r: &[u8]) -> Result<ResponderIdentity, EDHOCError> {
+    if r.is_empty() {
+        Ok(ResponderIdentity::Psk)
+    } else {
+        // A present `r` means the Python caller wants the stat-stat responder identity.
+        let mut identity = BytesP256ElemLen::default();
+        identity.copy_from_slice(r);
```
WilliamTakeshi: > This function will panic if the two slices have different lengths


copy_from_slice will panic if we send a r that is not r.is_empty() or r.len() == 32.

So, if the python user create something wrong, it will have an `panic!`. Can you `.try_into().map_err(|_| EDHOCError::ParsingError)?`?

lakers-python/src/responder.rs
```diff
@@ -105,15 +113,23 @@ impl PyEdhocResponder {
         let ead_2 = ead_2.try_into()?;
         let mut r = BytesP256ElemLen::default();
         r.copy_from_slice(self.r.as_slice());
```
WilliamTakeshi: I think now `self.r` can be `[]`, so it will panic on `copy_from_slice` (it will panic on every PSK call??)

This is not your fault, since the panic! was already there, but we always send r.len() == 32. Now that we can have different `r`, this code breaks.

I think you have to run this code only on StatStat `match` branch (or refactor how the python API is done) :((

lakers-python/src/initiator.rs
```diff
@@ -128,24 +136,37 @@ impl PyEdhocInitiator {
...
+        // stat-stat needs the static DH private key, PSK does not.
+        let identity = match self.start.method {
+            EDHOCMethod::StatStat => {
+                let i = i.ok_or_else(|| {
+                    PyValueError::new_err("StatStat requires initiator private key I")
+                })?;
+                InitiatorIdentity::StatStat {
+                    i: i.as_slice()
+                        .try_into()
+                        .expect("Wrong length of initiator private key"),
```
WilliamTakeshi: You should never panic on the python FFI. its undefined behavior, you have to return the Err.

lib/src/edhoc.rs
```diff
@@ -931,6 +1067,35 @@ fn encrypt_decrypt_ciphertext_2(
...
+    crypto: &mut impl CryptoTrait,
+    prk_3e2m: &BytesHashLen,
+    th_3: &BytesHashLen,
+    ciphertext_3a: &BufferCiphertext2,
+) -> BufferCiphertext2 {
+    // convert the transcript hash th_2 to BytesMaxContextBuffer type
+    // let mut th_3_context: BytesMaxContextBuffer = [0x00; MAX_KDF_CONTEXT_LEN];
+    // th_3_context[..th_3.len()].copy_from_slice(&th_3[..]);
+
+    // KEYSTREAM_3 = EDHOC-KDF( PRK_2e,   0, TH_2,      plaintext_length )
```
WilliamTakeshi, line 1057: These comments are necessary?

lib/src/edhoc/psk.rs
```diff
@@ -0,0 +1,245 @@
...
+            id_cred: id_cred_psk.as_encoded_value(),
+            cred_r: cred_r.bytes.as_slice(),
+            ead_3: ead_3,
+        },
+    );
+    Ok(PreparedMessage3 { message_3, th_4 })
+}
+
+fn recover_cred_i_from_id_cred_psk(id_cred_psk: &IdCred) -> Result<Credential, EDHOCError> {
+    if let Some(cred_i) = id_cred_psk.get_ccs() {
```
WilliamTakeshi, line 257: get_ccs() doesn't work for `CredentialTransfer::ByReference`, so it will answer as MissingIdentity in this case. is this right?

lakers-c/src/initiator.rs
```diff
@@ -8,20 +8,20 @@ use crate::*;
 
 /// structs compatible with the C FFI
 
-#[derive(Debug)]
 #[repr(C)]
 pub struct EdhocInitiator {
+    pub method: EDHOCMethod,
```
WilliamTakeshi: The method was added here, but also inside start, wait_m2, etc...

is necessary to have on both?

WilliamTakeshi (commented):
There is a binary committed on `examples/lakers-c-native/lakers_c_native_psk`. Probably committed by mistake.

Its getting really good :))

lakers-python/src/responder.rs
ElsaLopez133: True, I did not think of this. Maybe we could make the arguments keyword only? This way postitional argument order is no longer relevant?

lakers-python/src/responder.rs
ElsaLopez133: I moved the r conversion into the EDHOCMethod::StatStat branch and call r_prepare_message_2 inside the match. That way StatStat still validates self.r as exactly 32 bytes and PSK never touches self.r

lib/src/edhoc.rs
ElsaLopez133: Probably no (old code)

lib/src/edhoc/psk.rs
ElsaLopez133: MissingIdentity was the closest existing error I found. When ID_CRED_I is sent by reference, for example as a kid, parsing succeeds but the code cannot recover CRED_I from the message alone. The resolver then fails with: “I need CRED_I, but I do not have it locally”, which I thought maps reasonably to MissingIdentity. If we want more precision, we could introduce a dedicated error such as CredentialResolutionFailed.

lakers-c/src/initiator.rs
ElsaLopez133: You are right. The method in EdhocInitiator is redundant, since it is already included in the states.

malishav:
Where are we at with this PR? I still see TODOs in the PR description. @ElsaLopez133 can you give us an update?

ElsaLopez133:
I believe the PR is ready to review. Here is a concise summary of what was done:
- `edhoc.rs` is kept as the main EDHOC protocol file and depending on the method we branch to different functions that are stored in `edhoc/statstat.rs` and `edhoc/psk.rs`.  Therefore,  `edhoc/statstat.rs` and `edhoc/psk.rs` contain only the parts that differ between the Static-Static and PSK methods.
- These method-specific files return result structs back to `edhoc.rs`, so the central flow can continue without duplicating the shared protocol logic.
- We store method-specific fields in `enums` such as `WaitM3MethodSpecifics`, `ProcessingM2MethodSpecifics`,` ProcessedM2MethodSpecifics`, and `ProcessingM3MethodSpecifics`. These enums allow the common state structs to carry only the extra data required by the selected method. The use of  `enums`  was chosen to avoid having `Option` values.

What would need to be done is to add the Responder logic in the C bindings, but this would be a separate PR I believe (as it is neither there for the stat-stat protocol flow in the main branch)

malishav:
> I believe the PR is ready to review. Here is a concise summary of what was done:
> 
> * `edhoc.rs` is kept as the main EDHOC protocol file and depending on the method we branch to different functions that are stored in edhoc/statstat.rs`and`edhoc/psk.rs`. Therefore,  `edhoc/statstat.rs`and`edhoc/psk.rs` contain only the parts that differ between the Static-Static and PSK methods.
> * These method-specific files return result structs back to `edhoc.rs`, so the central flow can continue without duplicating the shared protocol logic.
> * We store method-specific fields in `enums` such as `WaitM3MethodSpecifics`, `ProcessingM2MethodSpecifics`,` ProcessedM2MethodSpecifics`, and `ProcessingM3MethodSpecifics`. These enums allow the common state structs to carry only the extra data required by the selected method. The use of  `enums`  was chosen to avoid having `Option` values.
> 
> What would need to be done is to add the Responder logic in the C bindings, but this would be a separate PR I believe (as it is neither there for the stat-stat protocol flow in the main branch)

Can you update the PR description with this text? thanks

WilliamTakeshi (approved):
LGTM

@chrysn , when you have time, can you take a look? I think we can merge this one and slowly improve some parts

PR #423, opened 2026-06-23
Final verifications for shared

WilliamTakeshi:
Recommend reviewing commit by commit
Continuation of https://github.com/lake-rs/lakers/pull/421

Add verification for Crypto + Helpers + ConnId + EadItems + EadItem + Edhoc_parser

fixed bugs:

Latent overflow in ConnIdType::length() for ByteString variant
Panic in ConnId::from_decoder() when byte string exceeds buffer capacity
Overflow in parse_single_ead() on i32::MIN label via .abs()

### DIFF
--- a/shared/src/cred.rs
+++ b/shared/src/cred.rs
@@ -1,8 +1,12 @@
 use super::*;
 
-pub type BufferCred = EdhocBuffer<192>; // arbitrary size
-pub type BufferKid = EdhocBuffer<16>; // variable size, up to 16 bytes
-pub type BufferIdCred = EdhocBuffer<192>; // variable size, can contain either the contents of a BufferCred or a BufferKid
+pub const BUFFER_CRED_LEN: usize = 192; // arbitrary size
+pub const BUFFER_ID_CRED_LEN: usize = 192; // variable size, can contain either the contents of a BufferCred or a BufferKid
+pub const BUFFER_KID_LEN: usize = 16; // variable size, up to 16 bytes
+
+pub type BufferCred = EdhocBuffer<BUFFER_CRED_LEN>;
+pub type BufferKid = EdhocBuffer<BUFFER_KID_LEN>;
+pub type BufferIdCred = EdhocBuffer<BUFFER_ID_CRED_LEN>;
 pub type BytesKeyAES128 = [u8; 16];
 pub type BytesKeyEC2 = [u8; 32];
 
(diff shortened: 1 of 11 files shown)


shared/src/cred.rs
```diff
@@ -140,24 +148,27 @@ impl IdCred {
...
-        [IdCredType::KID].contains(&self.item_type())
+    #[hax_lib::requires(self.bytes.len() >= 2 && self.bytes.len() <= BUFFER_ID_CRED_LEN)]
+    pub fn reference_only(&self) -> Result<bool, EDHOCError> {
+        Ok(self.item_type()? == IdCredType::KID)
     }
 
-    pub fn item_type(&self) -> IdCredType {
-        self.bytes.as_slice()[1].into()
+    #[hax_lib::requires(self.bytes.len() >= 2 && self.bytes.len() <= BUFFER_ID_CRED_LEN)]
+    fn item_type(&self) -> Result<IdCredType, EDHOCError> {
```
chrysn, line 157: With https://github.com/cryspen/hax/pull/2047 closing https://github.com/cryspen/hax/issues/899, it might be possible again to put constraints on the type, and then have the item having a valid IdCredType as a type invariant of Self.

karthikbhargavan:
can you collect all hax blockers in a document so we can address them? we
are keen to make it work for lakers.

abentkamp:
Hello! @karthikbhargavan asked me to look into the issues that you had to work around here. I noticed that some of them are already resolved on the main branch of hax. For others, I created new issues on the hax repo. I'll have to see if there are any that we can easily resolve.

Here is an overview:
https://docs.google.com/spreadsheets/d/1mwdRMcVtDbIbPTx7lOO7iRaqbdKF2axYlCK75TitIJs/edit?usp=sharing

WilliamTakeshi:
Hello @abentkamp I was planning to take a try on the new hax backend using lean+aeneas. Maybe it has more features.

Or do you think its best to continue working using F*?

abentkamp:
The new Lean backend is our focus right now, so it would be great to get feedback from you what works and what doesn't. At the moment, the Lean backend still has a few rough edges more than the F* backend, but I anticipate that this will change soon. We'll invest more effort in fixing Lean issues than in fixing F* issues, so switching might be worth it.

Right now, you should probably work with what's on our main branch because a lot has changed since the last release. Have a look at the tutorial to see how it works (it also assumes you use the main branch). Let me know how it's going.

abentkamp:
Heads up: We just changed how we deploy our tutorial. It is now based on the latest release. If you'd prefer to use `main`, follow the tutorial here: https://github.com/cryspen/hax/blob/main/docs/manual/lean/tutorial/panic-freedom.md

PR #425, opened 2026-07-08
Update example for nRF52840

WilliamTakeshi:
I was writting examples for the STM32 and nRF54 (I am experimenting integrating with `embedded-cal`). So I used nRF52840 as an example.

But seems like this example is outdated, so this PR is a work after me to update the API usage and remove all warnings.

Tested on 2 nRF52840-DK


WilliamTakeshi:
I will merge this one, since it only touches the example (that was broken)

PR #426, opened 2026-07-10
Add embedded-cal crypto backend + hardware-accelerated nRF54L15 example

WilliamTakeshi:
Adds a 4th lakers crypto backend built on the embedded-cal Cryptographic Abstraction Layer, letting lakers run on hardware crypto accelerators. Wires up dispatch features for the nRF54L15 (CRACEN) and STM32WBA55, and ships a runnable bare-metal nRF54L15 EDHOC example (initiator + responder) that runs the CRACEN end-to-end.

Disclosure: The radio code inside examples was generated by AI (Claude). Since embassy didn't provide an radio module, I generated one naive version.


WilliamTakeshi:
rebased and fixed after we merged PSK. tested again with nRF54

crypto/Cargo.toml
```diff
@@ -22,6 +22,11 @@ lakers-crypto-cryptocell310 = { workspace = true, optional = true }
 lakers-crypto-rustcrypto = { workspace = true, optional = true }
 rand_core = { version = "0.6.4", optional = true, default-features = false }
 
+lakers-crypto-embedded-cal = { workspace = true, optional = true }
+embedded-cal-software-demo = { git = "https://github.com/lake-rs/embedded-cal", optional = true }
+embedded-cal-nrf54l15 = { git = "https://github.com/lake-rs/embedded-cal", optional = true }
+embedded-cal-stm32wba55 = { git = "https://github.com/lake-rs/embedded-cal", optional = true }
```
chrysn, line 28: Tripped me up, but https://github.com/lake-rs/lakers/pull/428 fixes that.

examples/lakers-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+    ParsingError,
+    TimeoutError,
+    RadioError,
+}
+
+pub struct Packet {
+    // total length that gets transmitted over the air, equals length of pdu + 1, for pdu_header
+    pub len: usize,
+    // 1-byte application-level header, used for filtering the packets
+    pub pdu_header: Option<u8>,
```
chrysn: If construction from a slice fails anyway for 0-length, why is this an option at all?

Reading later, as_bytes copies internally … this looks awefully complex for something that could just be a heapless slice with known semantics of the first byte. (Also, this doesn't need to be pub.)

examples/lakers-nrf54l15/README.md
```diff
@@ -0,0 +1,54 @@
...
+(the `lakers-crypto/nrf54l15` dispatch feature).
+
+## Binaries
+
+| Binary              | What it does                                                                              |
+| ------------------- | ----------------------------------------------------------------------------------------- |
+| `initiator`         | EDHOC Initiator: runs the full handshake (message_1 → message_4) over the raw radio.      |
+| `responder`         | EDHOC Responder: waits for message_1 and completes the handshake, looping to serve again. |
+| `edhoc`             | Single-chip demo: crypto self-tests + an EDHOC `prepare_message_1` on hardware crypto.    |
+| `test_crypto_trait` | Stepped diagnostic that exercises each hardware crypto op in isolation.                   |
```
chrysn: Those don't exist here. Copy-paste from the other nRF example?

crypto/lakers-crypto-embedded-cal/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+    ) -> EdhocBuffer<N> {
+        let alg = aes_ccm_algorithm::<C, Tag>();
+        let mut outbuffer =
+            EdhocBuffer::<N>::new_from_slice(plaintext).expect("plaintext fits the output buffer");
+        let aead = self.cal.aead();
+        let key = aead.load_from_keydata(alg, key);
+
+        #[allow(
+            deprecated,
+            reason = "EdhocBuffer has no non-deprecated mutable-slice accessor"
```
chrysn, line 124: I briefly considered whether we should then add one, but that module is hax checked and getting that mutable slice would not be supported in hax or in F*.

Maybe that'd be better with lean? (But we won't rewrite that module here, so it's fine here.)

crypto/lakers-crypto-embedded-cal/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+            .expect("sha-256 output is exactly 32 bytes")
+    }
+
+    type HashInProcess<'a>
+        = sha2::Sha256
+    where
+        Self: 'a;
+
+    #[inline]
+    fn sha256_start<'a>(&'a mut self) -> Self::HashInProcess<'a> {
```
chrysn, line 70: Can this not be implemented in terms of embedded-cal?

chrysn (commented):
A few comments and questions popped up.

As discussed earlier today, I've added two commits that change from "build an instance, trust it to be dropped, but also be glad that the nrf-pac hands out multiple copies" to "we have one Cal and reuse that in the loop". That that worked without stealing / unsafe makes me wonder if things could go terribly wrong in the nrf implementation if two instances worked at the same time, and who'd be to blame when things go wrong then, but that's not to be addressed here.

examples/lakers-nrf54l15/src/lib.rs
WilliamTakeshi: Updated here:

[`9f72ea1` (this PR)](https://github.com/lake-rs/lakers/pull/426/commits/9f72ea12dbd3b5afeac8231462de769ddc3ed87a)

examples/lakers-nrf54l15/README.md
WilliamTakeshi: Cleanup here:

[`4f8b15f` (this PR)](https://github.com/lake-rs/lakers/pull/426/commits/4f8b15fd8379eb0c64cf0563c48834d598747cbc)

crypto/lakers-crypto-embedded-cal/src/lib.rs
WilliamTakeshi: I plan to start migrating using the new backend in hax (lean + aeneas). It looks promissing

crypto/lakers-crypto-embedded-cal/src/lib.rs
WilliamTakeshi: [`2ffc608` (this PR)](https://github.com/lake-rs/lakers/pull/426/commits/2ffc6088e80df230e4f9d757f76704cc2355e995)

updated to use embedded-cal. But needed to add a `panic!` for `Default`

WilliamTakeshi:
@chrysn updated addressing all points. Can you re-review it? thanks!

crypto/lakers-crypto-embedded-cal/src/lib.rs
chrysn: That's a good caveat, which I've taken up both in https://github.com/lake-rs/embedded-cal/issues/116 and in https://github.com/lake-rs/lakers/issues/437 -- but as things are, thanks, that's as good a fix we can do before later refactoring.

examples/lakers-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+#[derive(Debug)]
+pub enum PacketError {
+    PayloadTooLong,
+}
+
+/// The radio DMAs straight into and out of this buffer, so it is laid out exactly as the frame
+/// goes over the air: the S0 byte, the LENGTH byte, then the PDU. The first PDU byte is the
+/// application-level header used to filter packets; the rest is the payload.
```
chrysn, line 45:
```suggestion
/// The radio DMAs straight into and out of this buffer, so it is laid out exactly as the frame
/// is sent to the hardware: the S0 byte, the LENGTH byte, then the PDU. The first PDU byte is the
/// application-level header used to filter packets; the rest is the payload.
/// For simplicity of the direct-to-hardware demo, it is required that that
/// CBOR encoded value has exactly one byte.
```

crypto/lakers-crypto-embedded-cal/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+impl<C> core::fmt::Debug for Crypto<C> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
+        f.debug_struct("lakers_crypto_embedded_cal::Crypto")
+            .field("cal", &core::any::type_name::<C>())
+            .finish()
+    }
+}
+
+impl<C: Cal + rand_core::TryCryptoRng> CryptoTrait for Crypto<C> {
+    fn supported_suites(&self) -> EdhocBuffer<MAX_SUITES_LEN> {
```
chrysn, line 50:
```suggestion
    fn supported_suites(&self) -> EdhocBuffer<MAX_SUITES_LEN> {
        // As support for different suites is added, this will be changed to query the CAL for for the algorithms it supports (possibly at creation time), and compute the suites from that.
        //
        // Right now, suite support is as hard-coded as the suites are hard-coded in the lakers Crypto trait (e.g. by having a sha256 function).
```

shared/src/crypto.rs
```diff
@@ -54,6 +54,9 @@ pub trait Crypto: core::fmt::Debug {
     /// hash.finalize().into()
     /// ```
     fn sha256_digest(&mut self, message: &[u8]) -> BytesHashLen;
+    /// This is in process of being moved to something more embeddec-cal friendly.
```
chrysn:
```suggestion
    /// This is [in process of being moved](https://github.com/lake-rs/lakers/issues/437) to something more embedded-cal friendly.
```

(This is not only for the users: I like having links to issues around where we have workarounds, so that when we address the issue, we can grep for its address and find where we already knew we'd have to change things).

chrysn (approved):
Thanks, LGTM.

I'd appreciate if the comment suggestions were applied before merging (or some variation on it or otherwise being processed), but I think this is good now.

Before we release, we'll want to have a released version of embedded-cal, but a) that's the plan, and b) if it turns out we want to do a release before, we can always just not release this particular crate.

chrysn:
Thanks.

What I forgot to write in the review: I did not test this for lack of hardware (blocked by https://github.com/lake-rs/embedded-cal/issues/112), but trust your testing.

Please merge.

WilliamTakeshi:
Tested on my hardware! Both before and after merge

PR #428, opened 2026-07-18
doc: Describe lakers-crypto as not being a go-to crate for applications

chrysn:
Reviewing https://github.com/lake-rs/lakers/pull/426 I was confused by the dependencies being added – the ones in there make sense for testing, but applications should not rely, actually should long not have relied on lakers-crypto but use their own judgement.

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -4,7 +4,7 @@ version.workspace = true
 edition = "2021"
 authors = ["Mališa Vučinić <malisa.vucinic@inria.fr>"]
 license.workspace = true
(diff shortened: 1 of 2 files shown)


WilliamTakeshi (approved):

malishav:
can we go ahead and merge this?

chrysn:
Yes.

We never talked much about contributing guidance in lakers, but if https://github.com/lake-rs/embedded-cal/pull/115 works for embedded-cal, maybe we'd like to essentially 1:1 use it here as well? That'd clarify that either you or William could have merged it, as soon as it was approved.

PR #429, opened 2026-08-11
Allow encode_info to have context > 255

WilliamTakeshi:
working with sig-sig, implementing the a test case for [Authentication with Signatures, X.509 Identified by 'x5t'](https://datatracker.ietf.org/doc/html/rfc9529#name-authentication-with-signatu) I saw that MAC_2 needed a `context > 255`, and it fails on the code that we have today. 

2 bytes should be enough, since `MAX_KDF_CONTEXT_LEN` goes up to 1024 via the `max_kdf_content_len_*` features.


malishav:
resolve conflict then merge?

PR #431, opened 2026-08-12
Implement method 0 (SigSig)

WilliamTakeshi:
PR is on top of #430

Recommended reviewing by commit, since its a bit too long 

This PR implements EDHOC method 0 (SigSig).

* It starts adding ECDSA-P256 sign/verify for the trait and all 3 crypto backends
* Then renames `statstat.rs` to `stat.rs`
* Creates a `sig.rs` that mirrors `stat.rs`
* Add new variants to some enums 
* wire everything together in `edhoc.rs`
* create a handshake full test for sigsig
* also add some tests based on the [testvector of RFC9529](https://datatracker.ietf.org/doc/html/rfc9529#name-authentication-with-signatu)

This PR doesn't implement the C and python bindings. If merged, create a issue to add it

## Disclaimer

Some of this code was generated by AI, specially the part that implements the backend for both PSA and Cryptocell. I revised everything manually and tested everything


malishav:
In the interest of facilitating a review of this PR, could you split out all crypto-related changes?

paternolli:
Thanks for this — having method 0 in lakers is great to see.

One question on the signing API before it settles. `p256_ecdsa_sign(&mut self,
private_key: &BytesP256ElemLen, ...)` requires raw private-key bytes in-process,
which prevents backends from representing persistent/non-exportable signing
keys such as PSA key IDs, PKCS#11/HSM/TPM keys, or OS keystores.

Would you consider making the signing-key type backend-defined, e.g.
`type SigningKey;` with
`sign(&mut self, key: &Self::SigningKey, message: &[u8])`?

Existing byte-oriented backends could keep `BytesP256ElemLen`, while
handle-based backends could use their own key-reference type. This also seems
aligned with embedded-cal's handle-oriented key model.

I also noticed the shared deterministic-signing test may be stronger than
EDHOC/ES256 requires; verification seems like the backend-neutral property.

Happy to move this to a separate issue or help with a patch if useful.

PR #433, opened 2026-09-03
Add EDHOC resumption PSK derivation

ElsaLopez133:
This PR adds initial support for deriving EDHOC resumption material

Changes:
- Add resumption PSK constants and `BytesResumptionPsk`.
- Add `ResumptionPsk` result type containing `rPSK`, `rKID`, and `rID_CRED_PSK`.
- Add `IdCred::from_kid()` to construct `{ 4 : kid }` from raw kid bytes.
- Add `derive_resumption_psk()` for completed initiator and responder sessions.
- Add a test checking both peers derive matching resumption material.

Limitations:
- This PR only derives resumption material.
- A resumed PSK exchange is not wired yet because current PSK credentials are fixed to 16-byte symmetric keys, while `rPSK` is derived at hash length. This is fixed in PR #434

### DIFF
--- a/lib/src/lib.rs
+++ b/lib/src/lib.rs
@@ -868,46 +868,6 @@ mod test {
 
         assert_eq!(i_oscore_secret, r_oscore_secret);
         assert_eq!(i_oscore_salt, r_oscore_salt);
-    }
-    #[cfg(feature = "test-ead-none")]
-    #[test]
-    fn test_resumption() {
-        let cred_i = Credential::parse_ccs_symmetric(CRED_I_PSK.try_into().unwrap()).unwrap();
-        let cred_r = Credential::parse_ccs_symmetric(CRED_R_PSK.try_into().unwrap()).unwrap();
-
-        let initiator =
-            EdhocInitiator::new(default_crypto(), EDHOCMethod::PSK, EDHOCSuite::CipherSuite2);
-
-        let responder =
-            EdhocResponder::new(default_crypto(), ResponderIdentity::Psk, cred_r.clone());
-
-        let (initiator, message_1) = initiator.prepare_message_1(None, &EadItems::new()).unwrap();
-
-        let (responder, _c_i, _ead_1) = responder.process_message_1(&message_1).unwrap();
-        let (responder, message_2) = responder
-            .prepare_message_2(CredentialTransfer::ByReference, None, &EadItems::new())
-            .unwrap();
-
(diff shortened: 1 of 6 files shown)


ElsaLopez133:
TODO: The error comes form the PSA backend which cannot allocate another volatile key during test::test_resumption.

lib/src/lib.rs
```diff
@@ -862,6 +870,32 @@ mod test {
...
+                CBOR_MAJOR_MAP + 1,
+                KID_LABEL,
+                CBOR_MAJOR_BYTE_STRING | RESUMPTION_PSK_KID_LEN as u8,
+            ],
+        );
+
+        assert_eq!(
+            &i_resumption.rid_cred_psk.as_full_value()[3..],
+            i_resumption.kid.as_slice(),
+        );
```
WilliamTakeshi, line 896: do we have also a testvector to add to the tests?

this test we kinda duplicated, since earlier we checked that both initiator and responder have the same data. So this are just re-running everything again on the same function twice and checking if they are eq.

lib/src/lib.rs
```diff
@@ -304,6 +304,10 @@ impl<Crypto: CryptoTrait> EdhocResponderDone<Crypto> {
     pub fn edhoc_key_update(&mut self, context: &[u8]) -> [u8; SHA256_DIGEST_LEN] {
         edhoc_key_update(&mut self.state, &mut self.crypto, context)
     }
+
+    pub fn derive_resumption_psk(&mut self) -> Result<ResumptionPsk, EDHOCError> {
```
WilliamTakeshi, line 313: Here is after we have the state `EdhocResponderDone`. But we can generate this state without checking message_4 (`fn completed_without_message_4(self) -> Result<EdhocResponderDone<Crypto>, EDHOCError>`).

Genuine question, since I am not familiar with PSK, on the earlier exchange, message_4 is mandatory? If yes, then what happens if we run completed_without_message_4 and then try to do a derive_resumption_psk??

ElsaLopez133: I fixed it by not allowing PSK method to not do a message 4 (since in PSK a method 4 is mandatory)

WilliamTakeshi: My understanding is that its not fully fixed:

this still possible on statstat, right?

```rust
let mut initiator = initiator.completed_without_message_4().unwrap();  // Ok
let r = initiator.derive_resumption_psk().unwrap();
```

PR #434, opened 2026-09-03
Allow variable length PSKs

ElsaLopez133:
Allow variable length PSKs (before it was hardcoded, but this is a problem when introducing resumption keys)

- Replace fixed-size PSK storage with BufferPsk.
- Parse CCS symmetric keys using the encoded CBOR byte-string length.
- Update RustCrypto, PSA, and cryptocell310 crypto backend trait implementations.
- Fix C wrapper credential conversion after CredentialKey stopped being Copy.

### DIFF
--- a/lib/src/edhoc.rs
+++ b/lib/src/edhoc.rs
@@ -5,7 +5,7 @@ mod psk;
 mod statstat;
 
 use psk::{
-    i_parse_message_2_psk, i_prepare_message_3_psk, i_verify_message_2_psk, r_parse_message_3_psk,
+    i_parse_message_2_psk, i_prepare_message_3_psk, i_verify_message_2_psk,
     r_parse_message_3_psk_with_cred_resolver, r_prepare_message_2_psk, r_verify_message_3_psk,
 };
 use statstat::{
@@ -179,9 +179,7 @@ pub fn r_parse_message_3(
         WaitM3MethodSpecifics::StatStat { .. } => {
             r_parse_message_3_statstat(state, crypto, message_3)?
         }
-        WaitM3MethodSpecifics::Psk { cred_r } => {
-            r_parse_message_3_psk(state, crypto, message_3, cred_r)?
-        }
+        WaitM3MethodSpecifics::Psk { .. } => return Err(EDHOCError::UnsupportedMethod),
     };
 
     Ok((
--- a/lib/src/edhoc.rs
+++ b/lib/src/edhoc.rs
@@ -1097,7 +1097,7 @@ fn compute_prk_4e3m(
 fn compute_prk_4e3m_psk(
     crypto: &mut impl CryptoTrait,
     salt_4e3m: &BytesHashLen,
-    cred: &BytesElemLenPSK,
+    cred: &BufferPsk,
 ) -> BytesHashLen {
     crypto.hkdf_extract_psk(salt_4e3m, &cred)
 }
@@ -1926,8 +1926,28 @@ mod tests {
 
     #[test]
     fn test_compute_prk_4e3m_psk() {
-        let prk_4e3m = compute_prk_4e3m_psk(&mut default_crypto(), &SALT_4E3M_PSK_TV, &PSK_TV);
+        let psk = BufferPsk::new_from_slice(&PSK_TV)
+            .map_err(|_| EDHOCError::ParsingError)
+            .unwrap();
+        let prk_4e3m = compute_prk_4e3m_psk(&mut default_crypto(), &SALT_4E3M_PSK_TV, &psk);
         assert_eq!(prk_4e3m, PRK_4E3M_PSK_TV);
+    }
+
+    // fn test_compute_prk_4e3m_rpsk() {
+    //     let prk_4e3m = compute_prk_4e3m_psk(&mut default_crypto(), &SALT_4E3M_PSK_TV, &R_PSK_TV);
+    //     assert_eq!(prk_4e3m, PRK_4E3M_R_PSK_TV);
+    // }
+    #[test]
+    fn test_symmetric_32_bytes() {
+        let mut psk_32 = BufferPsk::new();
+        let mut psk_16 = BufferPsk::new();
+        psk_32.fill_with_slice(&[0xAB; 32]).unwrap();
+        psk_16.fill_with_slice(&[0xAB; 16]).unwrap();
+
+        let prk_32 = compute_prk_4e3m_psk(&mut default_crypto(), &SALT_4E3M_PSK_TV, &psk_32);
+        let prk_16 = compute_prk_4e3m_psk(&mut default_crypto(), &SALT_4E3M_PSK_TV, &psk_16);
+
+        assert_ne!(prk_32, prk_16);
     }
 
     #[test]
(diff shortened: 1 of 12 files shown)


shared/src/cred.rs
```diff
@@ -292,18 +285,45 @@ impl Credential {
...
+                    if value.len() <= len_offset {
+                        return Err(EDHOCError::ParsingError);
+                    }
+
+                    (value[len_offset] as usize, psk_header_offset + 2)
+                } else {
+                    return Err(EDHOCError::ParsingError);
+                };
+
+            if psk_len == 0 || psk_len > MAX_PSK_LEN {
```
WilliamTakeshi: Here we are accepting psk_len [1, 32], should be >=16

> 9.5.  Cryptographic strength
> Each external PSK MUST be derived from at least 128 bits of entropy, and MUST be at least 128 bits long

Fix: enforce a floor

shared/src/cred.rs
```diff
@@ -273,16 +273,9 @@ impl Credential {
     pub fn parse_ccs_symmetric(value: &[u8]) -> Result<Self, EDHOCError> {
         const CCS_PREFIX_LEN: usize = 3;
```
WilliamTakeshi: AFAIU, this function `parse_ccs_symmetric` is skipping the whole start of the CCS

`CNF_AND_COSE_KEY_PREFIX_LEN = 8` and `COSE_KEY_ITEMS_BEFORE_PSK_HEADER_LEN = 2` skips the whole first 10 bytes.

We should make it more like `parse_ccs` and `parse_cosekey` to check things like is `kty = 4` or the user called the wrong function on this CCS, etc...

FIX: make it more like parse_ccs and parse_cosekey correctly with a streaming CBORDecoder, checking every label and calling finished()

shared/src/cred.rs
```diff
@@ -292,18 +285,45 @@ impl Credential {
...
+                };
+
+            if psk_len == 0 || psk_len > MAX_PSK_LEN {
+                return Err(EDHOCError::ParsingError);
+            }
+            let psk_end = psk_start
+                .checked_add(psk_len)
+                .ok_or(EDHOCError::ParsingError)?;
+            if psk_end > value.len() {
+                return Err(EDHOCError::ParsingError);
```
WilliamTakeshi: related with comment above (use streaming CBORDecoder)

fox example, if someone send a VALID_CRED_PSK_16 || TRASH_BITS, the function will gladly accept and keep these TRASH_BITS somewhere that they may be used

shared/src/cred.rs
```diff
@@ -292,18 +285,45 @@ impl Credential {
...
+            if value.len() <= psk_header_offset {
+                return Err(EDHOCError::ParsingError);
+            }
+
+            let header = value[psk_header_offset];
+
+            let (psk_len, psk_start) =
+                if header >= CBOR_MAJOR_BYTE_STRING && header <= CBOR_MAJOR_BYTE_STRING_MAX {
+                    (CBORDecoder::info_of(header) as usize, psk_header_offset + 1)
+                } else if header == CBOR_BYTE_STRING {
```
WilliamTakeshi: related with comment above (use streaming CBORDecoder)

If the bytestring > 24, its a bug here

shared/src/lib.rs
```diff
@@ -114,10 +114,10 @@ pub const CBOR_MAJOR_ARRAY: u8 = 0x80u8;
...
 pub const CBOR_MAJOR_MAP: u8 = 0xA0;
 pub const MAX_INFO_LEN: usize = 2 + SHA256_DIGEST_LEN + // 32-byte digest as bstr
-				            1 + MAX_KDF_LABEL_LEN +     // label <24 bytes as tstr
-						    1 + MAX_KDF_CONTEXT_LEN +   // context <24 bytes as bstr
-						    1; // length as u8
-
+    1 + MAX_KDF_LABEL_LEN +     // label <24 bytes as tstr
+    1 + MAX_KDF_CONTEXT_LEN +   // context <24 bytes as bstr
+    1; // length as u8
+pub const MAX_PSK_LEN: usize = SHA256_DIGEST_LEN;
```
WilliamTakeshi, line 124: Is there something binding MAX_PSK_LEN with SHA256_DIGEST_LEN?

or we just picked `MAX_PSK_LEN = 32` as an arbitrary number and happen to be also the same as SHA256?  I would just put something like this to the next dev to know they can change if they want to
```rust
pub const MAX_PSK_LEN: usize = 32; // arbitrary chosen number
```
or a comment explaining why they have to be the same, bc its not clear to me why :(

shared/src/lib.rs
```diff
@@ -174,6 +174,7 @@ pub type BufferMessage3 = EdhocMessageBuffer;
 pub type BufferMessage4 = EdhocMessageBuffer;
 pub type BufferCiphertext2 = EdhocMessageBuffer;
 pub type BufferCiphertext3 = EdhocMessageBuffer;
+pub type BufferPsk = EdhocBuffer<MAX_PSK_LEN>;
```
WilliamTakeshi: is this a secret?

because EdhocBuffer derives `Debug`, anyone with access to the code can see it just by printing it. (same for CredentialKey)

Maybe add a TODO: we have to check this. (also check zeroize)

lib/src/edhoc.rs
```diff
@@ -1097,7 +1097,7 @@ fn compute_prk_4e3m(
 fn compute_prk_4e3m_psk(
     crypto: &mut impl CryptoTrait,
     salt_4e3m: &BytesHashLen,
-    cred: &BytesElemLenPSK,
```
WilliamTakeshi, line 1100: I think you removed all uses of `BytesElemLenPSK` + `ELEM_LEN_PSK` + `BytesKeyAES128` (is this one a misname?)

I think we can remove it from the code

shared/src/cred.rs
```diff
@@ -6,10 +6,10 @@ pub type BufferIdCred = EdhocBuffer<192>; // variable size, can contain either t
 pub type BytesKeyAES128 = [u8; 16];
 pub type BytesKeyEC2 = [u8; 32];
 
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
```
WilliamTakeshi: CredentialKey is not copy anymore, but I think we assume it is on C side:

```rust
    pub unsafe fn copy_into_c(cred: Credential, cred_c: *mut CredentialC) {
        (*cred_c).bytes = cred.bytes;
        (*cred_c).key = cred.key;
        (*cred_c).kid = cred.kid.unwrap();
        (*cred_c).cred_type = cred.cred_type;
    }
```

Can you check whats the implication??

WilliamTakeshi (commented):
Thanks Elsa! Really excited that you are back!!

Let me know if you need clarification in any of those.

shared/src/lib.rs
ElsaLopez133: I chose it arbitrarily. For resumption, the PSK is associated to the hash function, and in this particular example it was 32 bytes.

shared/src/lib.rs
ElsaLopez133: I added a TODO, I will try to fix this later

WilliamTakeshi (approved):

